PR #4, opened 2019-03-01
Some build fixes

bernds:
This allows cmake to find the cuda libraries on my Gentoo system, and allows building in a separate build directory which I think is generally preferred.


tychota:
Confirmed, that is also the fix I need on gentoo.

lightvector:
Merged, thanks!

PR #17, opened 2019-06-04
Move *Desc structs to separate file; document & clean-up nninterface

lpuchallafiore:
Also, add .clang-format with default Chromium settings (closest of the defaults
to the current project style).


lightvector:
* Changed some of the clang format settings. Also, if you don't mind, I bumped the line length to 120 instead of 80. If you're in the same boat as me (I have a relatively wide screen compared to height and don't often do windows side-by-side), then even longer would be nice too. Formatting seems good now, but there are still some rough edges compared to the style elsewhere, so if possible please try not to reformat large chunks of existing files for now.

* I re-renamed some of the things you renamed. For one of the things, the ways you renamed them suggest maybe you weren't sure yet how it was used?

* Yay your added doc comments are very nice.

* Also pushed to master is the removal of support for versions before V3. Except for a tiny bit of post processing that you are free to punt on since it's all but trivial, there should be no difference at all between the remaining versions supported (V3,V4,V5,V6) in the neural net architecture, so you should not need to branch on version at all in the implementation (so long as you're parametrized over the integer constants for things like number of channels and height/width, which I can't imagine would be a problem).

PR #55, opened 2019-07-24
Make CMake neural net backend options mutually exclusive

TFiFiE:
Right now, if one calls CMake with either `-DUSE_OPENCL_BACKEND=1` or `-DUSE_CUDA_BACKEND=1` and later wants to switch to the other backend, one needs to explicitly clear the old option in the next CMake call or manually remove or manipulate `CMakeCache.txt` first.


lightvector:
Thanks!

PR #62, opened 2019-08-06
Use strtof when loading our model

anoek:
istream float parsing is notably slower than reading a word and
converting it to a float with strtof.

Using this in our model loading inner loop shaves about 20-30% of the
time it takes to load our model off of the disk.


lightvector:
Excellent, thanks for the observation and suggestion! I implemented version with a little bit of error checking and applied it to a few other locations as well (of course, none of them are anywhere near as contributing to the time as the convolution layers). Let me know if this works for you!

https://github.com/lightvector/KataGo/commit/43c34bdd2d6378a272cd7c74561459d3b9d32044

anoek:
Looks great :) :+1:

PR #77, opened 2019-10-06
SE Networks

lightvector:
This is an experimental branch that adds CUDA inference-time support for SE-like networks. It now should be quite easy to get SE network training by modifying the Tensorflow code (much easier than the CUDA), but unfortunately the CUDA implementation of it shows SE has quite significant overhead, so currently this branch on hold, as this doesn't seem promising without more optimization work.

So this pull request is definitely not ready to merge, and I've also stopped further work on it, but putting it up here anyways just to make it more visible that the possibility exists.


lightvector:
If we ever want this we'll probably just reimplement it, but it seems like global pooling already does what you'd want this to do.

PR #115, opened 2020-01-30
Fix compiler warning

dwt:
Just a small fix that clang warned about.


lightvector:
Thank you! I will incorporate this in next release.

PR #117, opened 2020-02-02
Fix for macOS

yenw:
This PR fixes the `unqualified-id` error on the macOS.


lightvector:
Yuck. Presumably macOS, unlike MSVC windows and g++ linux, globally defines macros called "TRUE" and "FALSE"?

yenw:
/usr/include/mach/boolean.h:81:17: note: 
      expanded from macro 'TRUE'
#define TRUE    1

yenw:
Perhaps replacing `enum value { True, False, AUTO };` with `enum value { False, True, AUTO };` is more appropriate.

lightvector:
This is actually merged in, although looks like GitHub got confused when I squashed all your commits into just one commit. Although conversion to integer is used nowhere I also implemented your suggestion of swapping the enum order, and I also changed AUTO to Auto to be consistent with capitalization. 

Thanks!

PR #129, opened 2020-02-13
Better benchmarking with ternary search + adjusting time/move

farmersrice:
Add a ternary search feature (through `-s` flag) that allows us to optimize between large thread ranges with significantly reduced computation. This will allow for much easier and quicker benchmarking on extremely powerful hardware. Control the maximum number of threads to search up to with `-m`. The higher the maximum, the more beneficial ternary search will be. 

(It's important to note that the ternary search will never be perfect, since the function of threads to strength isn't completely convex and has some noise built in. However, the estimate is fairly close to the maximum strength option. For small thread counts and moderate strength hardware, just testing manually with `-t` is probably good enough.)

The ternary search feature is also beneficial in that it provides a way to search for a good thread count and not just have the user tell the program to try a boatload of options.

Also added a time per move feature `-i` that should allow for more accurate strength estimates given a certain number of seconds per move of thinking time (this utilizes the existing formula). 

Here are some examples of outputs with these flags (irrelevant outputs omitted):
https://gist.github.com/farmersrice/f3a327945a4f95ac1ef595552e995f30

### DIFF
--- a/cpp/benchmark.cpp
+++ b/cpp/benchmark.cpp
@@ -354,17 +354,12 @@ int MainCmds::benchmark(int argc, const char* const* argv) {
                 << " avgBatchSize = " << Global::strprintf("%.2f",avgBatchSize)
                 << " (" << Global::strprintf("%.1f", totalSeconds) << " secs)";
 
-    cout << "\rnumSearchThreads = " << Global::strprintf("%2d",numThreads) << ":"
-         << " " << totalPositions << " / " << possiblePositionIdxs.size() << " positions,"
-         << " visits/s = " << Global::strprintf("%.2f",totalVisits / totalSeconds)
-         << " nnEvals/s = " << Global::strprintf("%.2f",numNNEvals / totalSeconds)
-         << " nnBatches/s = " << Global::strprintf("%.2f",numNNBatches / totalSeconds)
-         << " avgBatchSize = " << Global::strprintf("%.2f",avgBatchSize)
-         << " (" << Global::strprintf("%.1f", totalSeconds) << " secs)";
+    string outputString = outputStream.str();
+    cout << outputString;
     
     cout << std::flush;
 
-    return make_pair(eloEffect, outputStream.str());
+    return make_pair(eloEffect, outputString);
   };
 
   auto printResults = [&](vector<double> eloEffects, int bestEloEffectIdx) {
--- a/cpp/benchmark.cpp
+++ b/cpp/benchmark.cpp
@@ -8,6 +8,7 @@
 
 #include <chrono>
 #include <map>
+#include <sstream>
 
 #define TCLAP_NAMESTARTSTRING "-" //Use single dashes for all flags
 #include <tclap/CmdLine.h>
@@ -29,7 +30,7 @@ int MainCmds::benchmark(int argc, const char* const* argv) {
   string desiredThreadsStr;
   int numPositionsPerGame;
   bool useTernarySearch; // If enabled, we will ignore the desired threads parameter
-  int ternarySearchMax;
+  int ternarySearchMax = 32;
   int secondsPerGameMove;
   try {
     TCLAP::CmdLine cmd("Benchmark to test speed with different numbers of threads", ' ', Version::getKataGoVersionForHelp(),true);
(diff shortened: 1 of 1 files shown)


lightvector:
I like the general idea of testing adaptively to find a better number of threads, thanks.

Thoughts below:

> Add a ternary search feature (through `-s` flag) that allows us to optimize between large thread ranges with significantly reduced computation.
I'm not sure I understand "significantly reduced computation"? Like, nobody should ever be typing in and testing literally every possible number of threads from 1 to 64 if they care about the computation time. If we simply changed the default test to be slightly more intelligent, skipping low thread numbers on machines that appear to be very strong, then testing or searching within "8,12,16,24,32,48,64" would seem to also even cheaper, and simpler.

Maybe there's the oddball machine that works better with weird numbers of threads like "17" or "23" or something, but it seems to me like it's usually a waste of time to try to test such a fine granularity of number of threads. You're mostly just measuring noise at that point. Maybe ternary search within "1,2,3,4,5,6,8,10,12,16,20,24,32,40,48,64,80,96,..." would be better and much faster. (That particular pattern is the set of integers such that for every i, the set contains 2^i * {1,2,3,4,5,6}, deduplicated).

A disadvantage of the ternary search output, as implemented, is that it by presenting the results in a highly random order, it doesn't give the user a nice way to see the curve of visits per second or other statistics with respect to threads themselves.

> The ternary search feature is also beneficial in that it provides a way to search for a good thread count and not just have the user tell the program to try a boatload of options.

I think really we want users to also not have to specify the upper limit of the number of threads at all, but for that number to be determined dynamically, through something like successive doubling. And then if the method can be made good enough, it should replace the default behavior too, rather than having a separate flag.

Ultimately KataGo probably wants something like a config generator that adaptively writes/rewrites a config, automatically setting the number of threads, adjusting the rules, setting search limits, etc, based on asking the user a few simple questions, rather than requiring the user to edit a file. I want to think a bit more about how to make that work though.

farmersrice:
> Maybe ternary search within "1,2,3,4,5,6,8,10,12,16,20,24,32,40,48,64,80,96,..." would be better and much faster.

I think you're right; I did some more testing on 2x 2080ti instead of the 2070 used last time, and in addition to noise messing with the ternary search, a lot of these numbers that aren't multiples of powers of two are routinely slower than those that are. I've implemented the pattern you've mentioned; works much better. I've also noticed that you need really high visits to get an accurate benchmark; luckily ternary search helps with that! (Test results are [here](https://gist.github.com/farmersrice/85ccd0ea7d5583d7daff0300e4a2105b).) 

For some reason the visits/s on the ternary search mode are drastically different from those in the traditional mode, even at 20k visits, any idea why this might be? I haven't touched anything relating to the tree search or the neural net.

> A disadvantage of the ternary search output, as implemented, is that it by presenting the results in a highly random order, it doesn't give the user a nice way to see the curve of visits per second or other statistics with respect to threads themselves.

The latest commit prints out all the thread results once the ternary search finishes, so it should look alright now. 

> successive doubling

Added this. Now, we'll double the limit of the ternary search and repeat as long as the optimal number of threads found is > 2/3 of the search limit.

> And then if the method can be made good enough, it should replace the default behavior too, rather than having a separate flag.

I think it's probably better to have a separate flag and allow more users to test the ternary search on their own hardware first. If it's clear that the ternary search is significantly better then we can make it the default.

lightvector:
Sweet. I'll compile/test/merge this weekend. Thank you!

lightvector:
So, this has some messy complications where the batch size needs to be adjusted at the same time, otherwise the batch size will be too small as you perform successive doublings and such. But the batch size only gets set upon creation of the NNEvaluator, so that means we need to delete and recreate it.

Also, I think if we're going to have an auto-tuning thing, maybe we should split it off into a separate tool from this benchmark. It was already a suggestion of inbae from LZ Discord that we have an interactive configuration tool that actually writes out the config with various options (including things like choosing rules and the number of GPUs), rather than requiring the user to edit the config file by hand.

I'm working on this.

farmersrice:
Makes sense. Is there anything you would like me to work on?

lightvector:
There are a few things. Sent you a discord message. :)

PR #130, opened 2020-02-13
Document installation via home-brew

dwt:
Specifically documents where the networks and config files are installed.

What do you think? Should there be more documentation?

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ You can download a few selected neural nets from the [releases page](https://git
 ## Setting Up and Running KataGo
 KataGo implements just a GTP engine - GTP is a simple text protocol that Go software uses to communicate with engines. It does NOT have a graphical interface on its own. So generally, you will want to use KataGo along with a GUI or analysis program, such as [Lizzie](https://github.com/featurecat/lizzie) or [Sabaki](https://sabaki.yichuanshen.de/). Both of these programs also support KataGo's score estimates and visualization as well. NOTE: a version of KataGo is also packaged directly with Lizzie's latest Windows release, but this is a somewhat older version and doesn't support some recent features.
 
-KataGo currently officially supports both Windows and Linux. KataGo can be installed via [Homebrew](https://brew.sh) on MacOS (use `brew install katago`. The latest config files and networks are installed in katagos `share` directory. To find its location use `brew ls katago`). Generally, testing across different OS versions and compilers has not been done, so if you encounter problems, feel free to open an issue.
+KataGo currently officially supports both Windows and Linux. KataGo can be installed via [Homebrew](https://brew.sh) on MacOS (use `brew install katago`. The latest config files and networks are installed in KataGo's `share` directory. To find its location use `brew ls katago`). Generally, testing across different OS versions and compilers has not been done, so if you encounter problems, feel free to open an issue.
 
 ### OpenCL vs CUDA
 KataGo has both an OpenCL version and a CUDA version.
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ You can download a few selected neural nets from the [releases page](https://git
 ## Setting Up and Running KataGo
 KataGo implements just a GTP engine - GTP is a simple text protocol that Go software uses to communicate with engines. It does NOT have a graphical interface on its own. So generally, you will want to use KataGo along with a GUI or analysis program, such as [Lizzie](https://github.com/featurecat/lizzie) or [Sabaki](https://sabaki.yichuanshen.de/). Both of these programs also support KataGo's score estimates and visualization as well. NOTE: a version of KataGo is also packaged directly with Lizzie's latest Windows release, but this is a somewhat older version and doesn't support some recent features.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -81,7 +81,7 @@ You can download a few selected neural nets from the [releases page](https://git
...
 
-KataGo currently officially supports both Windows and Linux. KataGo should also run fine on OSX, but I don't personally have a good way to compile or test an OSX version - so for this you may need to compile it yourself or check with OSX users who have been able to compile it. Generally, testing across different OS versions and compilers has not been done, so if you encounter problems, feel free to open an issue.
+KataGo currently officially supports both Windows and Linux. KataGo can be installed via [Homebrew](https://brew.sh) on MacOS (use `brew install katago`. The latest config files and networks are installed in katagos `share` directory. To find its location use `brew ls katago`). Generally, testing across different OS versions and compilers has not been done, so if you encounter problems, feel free to open an issue.
```
lightvector: "katagos" -> "KataGo's"

Is there a way for the user to conveniently verify the config and network version, or is there different wording that the readme should use? For example, every time I push a new network or release, the "latest" will no longer be "latest" until you get around to updating it, right?

dwt: `brew info katago` will spit out the version of the packaging of katago - I guess that also `katago --version` should give that information. Networks and source sha256 are part of the package if that is what you mean with verification?

But yes, a package update of katago in home-brew is required to release either new networks or a new version of katago. Since in the past (as far as I have seen) those tended to often change in tandem, that would seem like a good idea anyway?

Regarding the package updates: I do not have ownership of that package in home-brew, i.e. everybody can do the pull request that changes the urls and sha256 sums to update the package. Does that help?

dwt:
It has been my experience that updating home-brew packages is far simpler than creating them anew, as much less knowledge of build-secrets is required for that (i.e. the original package creator usually figured that and the dependencies out already).

lightvector:
Are you intending that I should send a pull request to update the home-brew package every release? Or are you happy being the one to doing so, and the readme should say something like that the home-brew version is maintained by a volunteer ("dwt") and updated periodically, without claiming that it's always the latest version - which is likely to not be true right around release?

Release is already a bit of work, with the number of things that have to be updated, since it isn't well automated yet. I don't really want to add additional steps for myself here, where I don't actually personally have a convenient way to test on Mac.

Releases of the code and network versions are *not* expected to always come together, either may happen without the other. If things are done right, KataGo's executable should always be including its own hash, so that's fine. How is the model file named in the package you created? Is it named the same way it's done in the official releases ("g170-s....") or is there something that records what name it is? Otherwise, from a model file alone there's no practical way for a user to tell which one it is.

dwt:
Sorry, I wasn't clear - I am quite happy to create a pull request whenever I get to it. But there is nothing in principle that should hinder you from doing so yourself. Pull requests are built automatically by their build-infrastructure, so errors are usually easy to catch.

I surely don't want to heighten your workload unnecessary, you are already doing quite a bit for the community.

Regarding attribution: To be fair - the katago homebrew package is not maintained by me alone. This pull request for example was co-created by @apetresc, so surely not all of the blame / fame can go to me.

Regarding releases: Yes, model files retain their full name from your releases. I am not yet sure, but probably what I want to do is remove old models when new package versions / models are released - but I've yet to look into that.

I plan on adding a 'latest' symlink from a standard location that katago looks at, to facilitate auto choosing a network, if the `--network` parameter is omitted.

I'll try to reword my change request to reflect the external maintenance of the home-brew package better.

dwt:
Now that I think about it - I sure would be glad if @apetresc and me can become part of the contributors file. (Though I'm not sure about the process you have for that).

Speaking of contributions: I have reworded the pull request - do you think that it now better reflects how it should be?

apetresc:
> Regarding releases: Yes, model files retain their full name from your releases. I am not yet sure, but probably what I want to do is remove old models when new package versions / models are released - but I've yet to look into that.

I'm fairly certain Homebrew automatically removes files from the old Formula's manifest when a new one is installed; so as long as we keep the same name in the `resource` blocks (`15b-network` and `20b-network`) it should automatically replace the old ones.

> I plan on adding a 'latest' symlink from a standard location that katago looks at, to facilitate auto choosing a network, if the --network parameter is omitted.

I think this is a great idea even apart from the Homebrew implications :)

> It has been my experience that updating home-brew packages is far simpler than creating them anew, as much less knowledge of build-secrets is required for that (i.e. the original package creator usually figured that and the dependencies out already).

Indeed, for most Formulas, upgrades are just automated with `brew bump-formula-pr`,  and if the build-bot detects its just a URL+hash change on the same domain that passes CI it automatically approves it without human intervention. Our situation is a little bit more complicated because of the networks, but it should still be _significantly_ faster than getting it in in the first place.

Either way, no additional work required for you @lightvector :) Homebrew is largely maintained by people at arms-length away from upstream maintainers. Totally normal.

dwt:
Also, I'm glad to get pinged by you whenever a new release of katago happens to update the home-brew release.

lightvector:
This looks good, thank you! Also pushed a typo fix and minor readme edits and clarifications. and updated the CONTRIBUTORS file.

PR #131, opened 2020-02-13
[WIP] Standardize network argument and allow default paths if available

dwt:
Hi there,

as I mentioned before, I would like KataGo to have a default path it looks at to find a network that can be provided either by the user, or by a package (like the home-brew one).

I have to say that my C++ is _quite_ rusty, so I would certainly need some guidance on how to cleanly achieve this within the coding style of KataGo - but this is my current thinking how this could be achieved.

My current plan is that the function `defaultNNModelPath()` can take a look into the home directory of the user into something like `~/.katago/default_network.txt.gz`, then fall back to a package provided place that is configured at compile time or provide the empty default if none of these are available. Something like this?

Of course the type `KataGoCommandLine` needs to be extracted / renamed / moved to a central. location - but only if the direction I'm going here is actually something that you could endorse.

So, what do you think?

### DIFF
--- a/cpp/analysis.cpp
+++ b/cpp/analysis.cpp
@@ -38,13 +38,12 @@ int MainCmds::analysis(int argc, const char* const* argv) {
   int numAnalysisThreads;
   try {
     KataGoCommandLine cmd("Run parallel analysis engine");
-    TCLAP::ValueArg<string> configFileArg("","config","Config file to use (see configs/gtp_example.cfg)",true,string(),"FILE");
+    cmd.addConfigFileArg();
     cmd.addModelFileArg();
     TCLAP::ValueArg<int> numAnalysisThreadsArg("","analysis-threads","Analysis up to this many positions in parallel",true,0,"THREADS");
-    cmd.add(configFileArg);
     cmd.add(numAnalysisThreadsArg);
     cmd.parse(argc,argv);
-    configFile = configFileArg.getValue();
+    configFile = cmd.configFileArg.getValue();
     modelFile = cmd.modelFileArg.getValue();
     numAnalysisThreads = numAnalysisThreadsArg.getValue();
 
--- a/cpp/analysis.cpp
+++ b/cpp/analysis.cpp
@@ -5,10 +5,8 @@
 #include "program/setup.h"
 #include "program/playutils.h"
 #include "program/play.h"
+#include "commandline.h"
 #include "main.h"
-
-#define TCLAP_NAMESTARTSTRING "-" //Use single dashes for all flags
-#include <tclap/CmdLine.h>
 
 #include "external/nlohmann_json/json.hpp"
 
--- a/cpp/analysis.cpp
+++ b/cpp/analysis.cpp
@@ -37,16 +37,14 @@ int MainCmds::analysis(int argc, const char* const* argv) {
   string modelFile;
   int numAnalysisThreads;
   try {
-    TCLAP::CmdLine cmd("Run parallel analysis engine", ' ', Version::getKataGoVersionForHelp(),true);
+    KataGoCommandLine cmd("Run parallel analysis engine");
     TCLAP::ValueArg<string> configFileArg("","config","Config file to use (see configs/gtp_example.cfg)",true,string(),"FILE");
-    TCLAP::ValueArg<string> modelFileArg("","model","Neural net model file",true,string(),"FILE");
     TCLAP::ValueArg<int> numAnalysisThreadsArg("","analysis-threads","Analysis up to this many positions in parallel",true,0,"THREADS");
     cmd.add(configFileArg);
-    cmd.add(modelFileArg);
     cmd.add(numAnalysisThreadsArg);
     cmd.parse(argc,argv);
     configFile = configFileArg.getValue();
-    modelFile = modelFileArg.getValue();
+    modelFile = cmd.modelFileArg.getValue();
     numAnalysisThreads = numAnalysisThreadsArg.getValue();
 
     if(numAnalysisThreads <= 0 || numAnalysisThreads >= 16384)
--- a/cpp/analysis.cpp
+++ b/cpp/analysis.cpp
@@ -39,6 +39,7 @@ int MainCmds::analysis(int argc, const char* const* argv) {
   try {
     KataGoCommandLine cmd("Run parallel analysis engine");
     TCLAP::ValueArg<string> configFileArg("","config","Config file to use (see configs/gtp_example.cfg)",true,string(),"FILE");
+    cmd.addModelFileArg();
     TCLAP::ValueArg<int> numAnalysisThreadsArg("","analysis-threads","Analysis up to this many positions in parallel",true,0,"THREADS");
     cmd.add(configFileArg);
     cmd.add(numAnalysisThreadsArg);
--- a/cpp/benchmark.cpp
+++ b/cpp/benchmark.cpp
@@ -62,9 +62,8 @@ int MainCmds::benchmark(int argc, const char* const* argv) {
   bool autoTuneThreads;
   int secondsPerGameMove;
   try {
-    TCLAP::CmdLine cmd("Benchmark to test speed with different numbers of threads", ' ', Version::getKataGoVersionForHelp(),true);
+    KataGoCommandLine cmd("Benchmark to test speed with different numbers of threads");
     TCLAP::ValueArg<string> configFileArg("","config","Config file to use, same as for gtp (see gtp_example.cfg)",true,string(),"FILE");
-    TCLAP::ValueArg<string> modelFileArg("","model","Neural net model file to use",true,string(),"FILE");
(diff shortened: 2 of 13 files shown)


dwt:
The same could of course be done for other arguments. I would be quite fond of a standard `--version` argument for example. But that's something to discuss later.

lightvector:
Given that (unfortunately) right now the model file itself has no indication of which model it is, how do we propose to make that clear?

lightvector:
Note: The answer might be that I improve the model format. I've been wanting perhaps to make the model format into a binary one that is a bit faster to load. Float parsing is what makes the load time so long right now.

dwt:
I think meta data  in the model file would surely be very welcome. My current plan was to not put the model itself at the standard location, but instead symlink the original model file there. That way `redline()` or `ls -l` can always recover what model is actually used.

But that requires environment setup and not all users will do it that way, so meta data in the model file is _definitely_ a better way to do that.

If you are looking into binary formats for the model file, I'd definitely take a look at established formats. I did have good results with h5 in the past (easy to use, nice python libraries at least). But I guess that there are probably way more standard formats around now, and I would look towards the model interchange formats that IMO are part of stuff like [onnx](https://onnx.ai)

Apart from that: Do you think the way I structured the refactoring, to pull out common parameters, could fit the way you would like your code to be s structured?

lightvector:
I think the way you structured in terms of having a subclass like that seems very reasonable from a high level. Thank you!

As for details: does something different need to happen on Windows regarding the default path? Perhaps on Windows, looking within the same directory as the exe itself is more natural?

And apparently there was already a metadata field in the model format I created. But it wasn't being displayed, and also not populated quite the way it should. Hard to fix that for old model files even once I change to start displaying it now, but it's easy to fix for new model files and for a next release. Yay! Even with this, I still think the symlink is a good idea though.

I poked around at standard model formats. I'm fairly convinced now I should not attempt to use or support them. It would be a lot of complexity since they're all very general and have tons of metadata and structure that would be useless for me, since my GPU implementation only works for KataGo's exact NN structure rather than a general one, and so there's no point supporting a general format. The Tensorflow SavedModel weights are still there alongside the KataGo-specific format for people who want to do more general stuff.

dwt:
Regarding File formats: No surprise that the standard formats are complicated. Supporting them would IMO only make sense if you do not have to write a parser yourself, i.e. a dependency implements that completely.

If you want to go for a simpler binary format, as I said above, I'd have a look at hdf5 (or h5). There's good [python support](http://www.h5py.org) and [the format has many supporting libraries](https://en.wikipedia.org/wiki/Hierarchical_Data_Format). At least in python it did behave mostly as a drop in replacement for pickle when I used it.

Regarding the refactoring: Thanks for the endorsement, then I can continue this path and pull stuff out.

Regarding Windows support: Sadly I know next to nothing about windows programming. My plan would be: Support a default location on linux/Mac first, then if that works  a knowledgeable Windows user can hopefully easily contribute the same functionality there.

I was thinking that it may be nice to have a new command `katago set-standard-model /path/to/model.tar.gz' that would set up the symlink? Not sure this is a good idea, but may be a nice way to document this feature and make it accessible? Homebrew could then just call this command as part of the installation.

dwt:
I've had another go at expanding support for a standard network location (still just for the gtp subcommand).

Dunno, what do you think of it? If you approve, I can start rolling this out to all the command line parsers?

lightvector:
Sorry for the delay in looking at this. I think this looks very reasonable. I'm definitely happy about the refactor of the command line parser. I think that can and perhaps should be done separately from this default network path stuff. I'm willing to just accept that refactor immediately.

For the default network - maybe we should prepare for it to have a ".bin.gz" extension, or be able to handle both? Like look for .bin.gz first and then look for .txt.gz or something? With new versions we're about to release .bin.gz files instead which use a binary format for weights that is both smaller on disk and faster to load.

Also, if there isn't already, KataGo should definitely also improve its error message for if this default path doesn't exist, as well as to log what path it's reading from if it does exist and gets used, so we don't get any confusion over where it's actually looking on any instance. Also, is the only way you expect this path to be populated to be by package manager? It feels like the user's typically not going to know to set up a ".config/katago" themselves, that's even more complex than passing a different command line argument, which many users already do simply to have setups that use different neural nets (e.g. sometimes a big expensive one, sometimes you one of the smaller nets for speed or testing).

Windows support is necessary. For Windows, I think the least confusing would indeed just be to look in the directory of the executable, if that's possible. If not, then the current working directory may also be fine.

dwt:
Hi, I was alle to work on this a bit more - what do you think?

What I did was:
- Centralize the import and configuration of tclap
- Use the KataGoCommandLine object everywhere
- Extract a second centralised parameter 'configfileArg'

What do you think about that direction?

Next a standard path for the config could also be set and then I think I'm about done.

For the windows-specific implementation: I have no knowledge of windows - I can provide a default implementation for windows that just does not have default paths for those arguments. That would allow us to wait for a windows contributor who knows how it should be done on that platform to contribute it. What do you think about that?

dwt:
Oh, and I didn't extract https://github.com/lightvector/KataGo/blob/master/cpp/write.cpp#L1031 as that usage looked quite different than the other parsers? Maybe you could tell if that is intentional and should stay separate, or be unified too, to carry the same version information as all the other parsers.

dwt:
Ah well, of course documentation is something that is still needed - and maybe a command to promote a config / model file combo to the default config.

lightvector:
If you submit a version of this pull request that purely does the refactoring of the command line arguments, I am happy to accept it pretty much immediately, although I may myself do a slight bit of cleanup on top of that. Then we can think about the Windows case.

I don't think sticking something in the home directory for Windows is necessarily a good idea. At least, I find that in the case of Windows, when a program starts writing stuff into some hidden subdirectory of your home directory, or some AppData or LocalData directory, or whatever, discoverablility of this is hard, and it's much easier if the program just uses whatever's in the directory that contains the executable itself.

lightvector:
Or if you'd prefer, I can do the work to extract out that part from your code so far (along with additional cleanups) and then you can rebase.

dwt:
Sure, go ahead and take what you like - you know best what kind of changes you would want, and my C++ is really not up to snuff sadly.

I'll see that I rebase this pull request on top of that and add the defaults from there.

Adding support for windows to grab networks and config from the current directory should be relatively simple? You seem not to like the idea to just retain the current behaviour on windows for now and wait for someone to provide a good idea and implementation?

Regarding the idea to grab config from the  'current directory': Usually when invoking tools from other tools (i.e. katago from Lizzy) especially when started from the UI, the current directory can be badly / not intuitively defined, so that may actually not be so useful especially on windows where most people probably use katago that way.

I think that documenting the path where katago would look in the parameter description is quite a good hint on how to use that feature, what do you think about that?

dwt:
I have added default path implementation and a beginning of documentation for those parameters,  also something I'd like to know some feedback about. :-)

lightvector:
Ah, maybe keeping just the current behavior of no default at all on Windows would be fine, let me think about it. :)

dwt:
Maybe on Windows defaulting to the enclosing folder of the katago binary might be a good default?

lightvector:
I incorporated most of your changes, with a bunch of minor revisions and such, into master here:
a3cddbd

Since there was enough minor change that there would be a lot of conflicts, I rebased and squashed your commit stack here and then made some more fixes on top of that:
* Not accessing the filesystem when constructing the command line. ONLY apply the defaults later when we actually try to get the data out of the command line, and ONLY if the default is needed. Previously, if there would be some sorts of permissions or file system errors to look up the default paths, I think it could cause KataGo from failing to even display the help. And it would stop KataGo from working even if one tried explicitly to specify the config and model since looking up the default paths was not conditional. Errors would probably be rare, but would also be very bad so we should guard against this.
* Implementing the windows side of things to look in the executable directory. (Although I haven't actually tested it).
* Supporting .txt.gz default models as well.
* Adding logging of the config and model files to several places. Having defaults makes it important to do so because it is now suddenly more important to have it be discoverable in logs what config file or model file was actually being used.
* Sharing some pre-existing code that I forgot I had already to look up the user's home directory.

Pushed here. Let me know what you think.

https://github.com/lightvector/KataGo/compare/master...dwt-default_network_path-rebased

Additionally, do you have further thoughts on just using the exe directory on Linux too? Certainly it's a little unconventional for Linux - it makes it more awkward to imagine KataGo inhabiting a dir with a ton of other binaries, but I imagine that's not actually too common in practice., but also would be friendly in some ways for when the exe is isolated as probably is common in current usage.

lightvector:
I'm happy enough with this, so I just merged it into master. If you think it's still better to also just use the exe dir for Linux as well, let me know.

dwt:
Sorry, I’ve been a bit pre-occupied at https://wirvsvirushackathon.devpost.com <https://wirvsvirushackathon.devpost.com/>

I’ll get back to katago as soon as things clear up a bit.

Cheers,
Martin

PR #132, opened 2020-02-15
Allow KataGo to use CPU

farmersrice:
Was trying to set up a one-playout katago on a CPU-only VPS only to find out that CPUs aren't allowed for some reason. I can't see any reason why we shouldn't allow KataGo to run on the CPU. I get 1 playout per second, which is pretty slow, but considering that KataGo 1 playout is stronger than Leela 11 at 10k playouts, it's far from unusable. Thoughts?


lightvector:
This needed also some more changes as well - to add some safeguards to the heuristic selection of which OpenCL device to use, and prevent a CPU device from being selected over a real GPU device. But this was only a few lines change too, so easy enough.

lightvector:
(aff0c3a7)

PR #142, opened 2020-02-25
Add Support for KGS

lightvector:
This stack of changes implements `kgs-rules` and `kgs-time_settings`, so that KataGo's gtp engine can handle KGS server rulesets and time control methods. I think it should work, but I haven't actually tested it on KGS yet.

Also, it clarifies a bit of the documentation on Chinese rules.


lightvector:
I did a quick test, setting up a KGS bot briefly for the first time. Seemed to work, so I'm happy with this.

PR #161, opened 2020-03-11
set config to 9x9

yfuka86:


yfuka86:
sorry for the mistake

lightvector:
Not sure what this is, but no prob. :)

PR #165, opened 2020-03-11
Run selfplay consecutively

iopq:
This is for people who are doing experiments with only one or two GPUs and don't want to bother to setup GCP

The way you're supposed to use  it is you copy the nohup.sh file whereever you want to run it, modify the paths to point to the correct places and plop in keeper1.cfg, selfplay1.cfg with the correct game number limits

it does cycling learning rates because that works the best in my case - I can actually sit around learning since I only have one GPU to test it with


iopq:
one issue I can see is if the net fails to promote, it's not going to move the games since the folder will already exist

iopq:
So maybe I should replace it with `rsync -a "$BASEDIR"/selfplay/* "$BASEDIR"/selfplay_old --remove-source-files` for that reason

python/selfplay/consecutive/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Suggested settings for learning
+
+In `shuffle.sh` you should edit the min rows to 10000 because we're going to be working with less data to start. Expand window per row should be set to 1, taper window exponent to 1. In train.py you should decrease the samples per epoch to 250000 since we will generate fewer games to have a smaller learning loop. In your own copy of `selfplay1.cfg` you should set the amount of games to 3000, keeping cheap visits on. I suggest you reduce the visits in `keeper1.cfg` down to 100 or lower, because it cuts into your selfplay time significantly.
+
+## Modifications for smaller board sizes.
+
+In both config files you should change the settings. Set `bSizes` to the board size you wish to train, `bSizeRelProbs` should be set to `1`. Set `komiAllowIntegerProb` to `1.0` to allow for draws. `handicapProb` should be `0.0` on boards smaller than 19x19 because it's too imbalanced.
```
lightvector: komiAllowIntegerProb defaults to 1.0 now and is not even present in the config file by default in the first place now.
handicapProb should still make sense on 13x13 and larger, particularly when handicapCompensateKomiProb is large (so that komi is set to offset it and make the game fair). Also I would think handicapCompensateKomiProb should even not be set to exactly 1, just so the bot has some minimal continual amount of data about massively imbalanced board positions and a way of learning how to play in those settings without going crazy. Score utility means there's still learning signal in those positions, for both policy and for value-related outputs.

lightvector:
Sorry for the delay - took a look at this now. Given that this just adds some new files, it seems okay, but do you want to re-think about the method of windowing and/or test some more, given your still-in-progress experimentation and possibly mixed results against g104?

Cyclic learning rate and the training method together makes this look quite unlike the normal AlphaZero training loop. And I don't understand yet why merely having 1 GPU should change fundamentally what method you use to suddenly have these things. 

The parallelism that you get from multiple GPUs - selfplays running at the same time as training and gating, etc -  is a *drawback*, not a benefit - because it means that you're not training on quite the latest data, rather you're training on a little out-of-date data always, etc. So it seems to me that if you only have 1 GPU, the baseline thing you might start with is simply the sequentialized version of the KataGo training - you have less compute, but per compute, sequential means you should already be more efficient. You could always of course still reduce visits and such to be a little faster at the start too.

But anything to do with cyclic learning rates and this thing of having a very small window size with no data overlap forever is very much an experimental idea on top of that, and it sounds like for you that experiment is still in progress, so I'm not sure about mixing that experimentation in, if it hasn't actually been compared in a controlled way to the more basic process.

Back when I only had 3 GPUs I found was able to reach top amateur levels in just a day or two - and even with only a partial number of KataGo's current training enhancements, I hadn't thought of or implemented them all at the time - with just the basic AlphaZero-like process (no cycling learning rate or special windowing or whatever). So at least it it shouldn't be strictly necessary to significantly change nature of the the training loop when compute is limited.

iopq:
The amount of games in the window is decided by the configuration file, which I didn't commit here. It seems like the amount of samples should be close to the number of rows in the shuffle.

For example, I had 5x the number of samples, and I learned three times with different learning rates on the same exact data (although I shuffled it in between, which may help). This was overfitting so it caused the rock-paper-scissors scenario. My key statistics for, say, `p1acc` went up TOO high.

I can comment out the cyclical learning rates for now, this part actually works since you want a learning rate cut every time you generate games (you get better networks that way). But anyone can disable/enable this part any time they want, of course

iopq:
I retested a few recent networks.

Head to head, my latest won 78% vs. the previous, while the previous also had over 65% winrate against the one before that.

I ran the head-to-head tests again and they came out the same (for example, 77% against the previous net). So then I ran the latest against two networks ago net to see if it's "rock-paper-scissors" or "evil twin". The result of that test was 73% winrate. It seems there's a bit of both. My latest is stronger against the network it was  trained against, but also considerably stronger than two networks ago.

I also ran them all against g170-668... and got statistically insignificant differences in strength. They were all pretty weak against it 20-28% and not monotonically increasing. Basically, the strength improvements against the main katago runs are invisible until you actually reach parity.

I'm going to decrease the samples to 250K and redo this test again - I have 200K rows of data so my recommendation to keep the samples close to the number of rows seems to be correct.

I want something like this: 60% pass against the previous network and 70% pass vs. two networks ago - this would indicate additive elo improvements (I just played around with an elo ratings calculator and apparently at these winrates it's actually still basically linear)

iopq:
With 250K samples and only two subepochs there's some kind of bugginess, the amount of learning is much lower than other settings, each subepoch is only 5 minutes. I'm not sure why, when I did 4 subepochs it was overall longer.

I had a 73% pass and a 62% pass. When I compared two versions back, the new version won 72%. The new version had a 33% win vs. g170-0.6G so it's the best winrate so far.

I'll set it to use the 200K window and report back.

iopq:
I tried with the same learning settings, but new candidates after a few improvements either have weak passes or stopped passing. I'm cutting the learning samples by 1/2 again to offset the extra learning and trying again

iopq:
Last pass was 51%, the previous network failed to pass, before that 52.6% pass.

Testing the previous network (failed one) vs. the newest network I got 49.6%, then three networks ago I got 50.8% pass (?!). When I tested against g170 0.6G with the older network I got 36% win

Basically no visible improvement

iopq:
I had some overlap from the previous day and it failed to promote even though I generated a lot of extra rows, the overlap was 10% at most and only won like 40% failing the promotion. Then next day it archived those games and learned on 200K fresh rows and promoted with 58%

With these exact settings (cycling learning rates, amount of samples in learning, number of games, etc.) I have a sweet spot. I imagine you just adjust the network size/playouts to get the velocity (how many networks per day) you want

iopq:
Tested the new net, 38% vs. g170 0.6G so nothing amazing, statistically doesn't look any better than the older one.

For what it's worth, you can kill -STOP $PID and kill -CONT $PID on the katago process as well. I was able to pause, run the gatekeeper vs. g170 and come back to selfplay. I will test the new net generated today vs. the "overlap" net and g170 and compare again

iopq:
Tested another net that passed with 73% - it got 37% vs. g170 0.6G

71% win vs. the overlapping net (two nets ago) so it's at least crushing its predecessors just fine

A note on the loss statistics, when it was overlapping it seemed to be overfitting, while doing it this way it doesn't - the p1acc only differs by 1% or so after I run my learning so it doesn't seem to overfit too much to the small window

iopq:
New net passed with 63%, 68% vs. two nets ago

39% vs. g170 0.6g - still basically statistically indistinguishable, but at least the scores are creeping up over time

iopq:
Naphthalin said it could be due to temperature. I increased the temperature to 1.2 and reduced the halftime to 9 for tomorrow

Today I had extra games due to the games resetting, so I had a 71% pass. I'm only testing vs. g170 now because it seems it passes vs. previous networks just fine by getting strong at one particular opening (although I'm not going to verify by going over the games, I'll just see if the temperature changes help)

iopq:
So the problem with my approach is the window is a bit small, I expanded the window to use the previous games (except for the games generated by the network right before this one) and it had almost a pass vs. g170

I'm going to set it to 1 million rows and reuse all of the previous games (that should go back 5 networks or so) in a more conventional setup

I also increased the temperature early game to 0.9 (although I don't know why there are two temperature settings?)

iopq:
so I might have overfitted to the old games, after initial boost, candidates stopped passing, so I reduced the learning and the window size so I get past those games more quickly

now doing 3100 games to get 50K rows, learning at 100K samples, 400K window size

seems I reached the point where my window size should get larger because my improvements with 200K slowed down - but they slowed down in a surprising way, with strong passes but no improvements vs. g170

iopq:
It had a dip in win rates after the window wasn't filled, I'll start tracking now since I'm running the 400K window experiment

I am at 37% win rate vs. g170 0.6g with 20b 0.16G

iopq:
36% with the next model, most recent one failed to pass

400K moving window doesn't seem like it's progressing either

iopq:
After 2 days of generating 400K fresh rows, it had 43% vs. g170

I'm okay with this rate of progress. Hopefully in two or four days I'll reach parity.

iopq:
47% just now for freshtwenty 0.174g

iopq:
0.196G beats g170 0.6G, now I test vs. the next network
40% vs. g170 1.04G
30% vs. hzy's 20b net

iopq:
0.199G is at 48% vs. 1.04G, not bad

PR #170, opened 2020-03-17
Add policy temperature annealing

farmersrice:
At the root of the search tree, start with `initialPolicyTemperature` as policy temperature. As you go deeper down the tree, the temperature approaches `finalPolicyTemperature`. This is exponential and the half life is specified with `policyAnnealingHalflife`. 

The key idea here is that at nearer to the top of the tree, you want to fan out and look for potentially surprising moves, while as you go deeper down the tree you want to think deeper and get closer to the end for a more accurate evaluation of the parent nodes. 

A test with settings:
```
   initialPolicyTemperature = 1.1
   finalPolicyTemperature = 0.9
   policyAnnealingHalflife = 6
```
on a 2080ti getting around 2200 visits/s, limited to 1 second per move, gives `rating difference: 11.28 +- 37.02 (95% confidence)` after 339 games (175 won, 164 lost). 

It will take a few thousand games to definitively establish that these settings are superior within the 95% bound. While these settings may not actually be superior, it's clear that there's a good possibility that they are. In addition, these settings can probably be tuned to find a much better optimum than what I've randomly guessed. (Currently the `initialPolicyTemperature` = `finalPolicyTemperature` = 1.0 by default (no change to current behavior).)

The caveat is that you add 8 extra bytes to each node, since when you're reusing the tree you'll have to rescale the policies appropriately, which requires knowing the previous iterations' temperature. This could be remedied with a clever trick or two.

The hope is that some extra tuning with better hardware than I have will give KataGo a respectable elo boost. I think an extra 20 elo is well within reach.


lightvector:
I understand the concept, but I'm skeptical about approaching the concept this way. Why should the policy temperature be a function of the depth of the tree rather than the number of visits a node has received? After all, as the visits at a node becomes very large, the marginal effect of each individual visit approaches zero, and therefore you should be more and more wanting to fan out your exploration with some of your visits just as you would at a shallower depth with less and less worry that the exploration visits will corrupt your evaluation. And when visits are low, that worry is significant with respect to the quality of evaluation of that particular node, regardless of whether you are at a low visits node at depth 4 or a low visits node at depth 10.

There is okay principled theory for allowing there to be a discontinuity in parameters at the root versus other nodes, because at the root the evaluation quality of the node itself never matters *even at low visits* unlike anywhere else in the tree. Because unlike deeper in the search the average winrate at the root node itself has no effect at all on any move choices. But the story seems much less clear for depth 2 versus 4 versus 10.

It seems to me that such a concept would be better explored as an adjustment to the PUCT formula and/or FPU values. Possibly the way the PUCT formula incorporates the policy term can be visit-count-dependent, and it seems likely there is a way to do so that although not literally equivalent to what you're trying here, should have much the same effect in the big picture.

Doing it as PUCT/FPU formula change should also have the benefit of:
* Not requiring any additional fields to be added to the nodes.
* Less non-local state to think about between searches (by "non-local state" I mean the kinds of state that interfere with tree-reuse, that need to change when the search resumes rather than being purely a function of the node itself and its subtree. The non-local state that exists is already is often a headache to think about and I'd strongly like to not add more.)
* Does not blow up the memory usage. The current implementation in this pull request I think might double the memory usage for the NNOutputs, since there will both be the temperature-modified version in the tree and the original version sitting in the NNCache, instead of just one shared version.

Also one should try to get much tighter bounds on that confidence interval for the Elo change. :)

farmersrice:
I can definitely agree that the memory usage needs to be a lot better, and the elo bounds too. 

> Possibly the way the PUCT formula incorporates the policy term can be visit-count-dependent, and it seems likely there is a way to do so that although not literally equivalent to what you're trying here, should have much the same effect in the big picture.

Hmm, but the PUCT formula already has the policy term contribution depending on sqrt(parent visits) / (1 + sqrt(child visits)), no? 



Also, just remembered this from the original paper:
>  this search control strategy initially prefers actions with high prior probability and low visit count, but asympotically prefers actions with high action-value.

So what you're suggesting would actually be the _opposite_ of the intended mechanism: instead play very sharply at few visits to get a better winrate, and once you've established a lot of visits feel free to branch out? That's an interesting thought.

lightvector:
Well, no, this is what *you* suggested. :)

Think of it this way: the ONLY way the policy affects the search is through the PUCT exploration. You're proposing applying a depth-dependent way to modify the policy. Therefore, you're proposing a way to modify the PUCT exploration at a node. Right?

Depth and visit counts are highly but not perfectly correlated and I'm thinking that depth is possibly wrong and visit count is possibly more correct, assuming the idea holds up at all. If that's the case, then now we seek a way to modify how the policy prior is incorporated into PUCT formula based on visits. It doesn't matter that there's already a scaling there based on visits. If your temperature idea is reasonable then it means we want another scaling beyond that. In particular, one that at low visits/depths (slightly) exponentially reduces P if P was small, but at high visits/depths (slightly) exponentially increases P if P was small.

farmersrice:
Yeah, I see it now. I'm thinking we can do something like replacing the current policy in the PUCT exploration as such: 



Seems like this would follow the general idea of the policy temperature annealing but for visits instead. Maybe flip the numerator and denominator to check the reverse effect as well.

Will try this soon. Would definitely be a lot more resource- and developer-friendly than policy temperature annealing.

lightvector:
Cool. 

The main difference that scaling the policy in the PUCT formula directly has different from doing the temperature at the neural net output level is that it's hard/impossible to normalize them to sum to 1.

Normalization to sum to 1 is not actually a thing that's strictly needed, it's merely that not summing to 1 behaves like a c_PUCT change. Like, if the policy probabilities summed to 2 instead, then it's exactly the same as them summing to 1 and c_PUCT being twice as large. So that's a thing to keep in mind. I don't know a-priori whether when you flatten the policy you also need to jointly adjust c_PUCT downward to compensate - which could be done, even if it would be hard/costly to get it exactly normalized. Or perhaps the "effective increase" (in some sense) to c_PUCT is desirable, in that the whole point of flattening the policy is to explore a bit more, but disproportionately on the lower-visited or unvisited children.

farmersrice:
Coming back to the original proposal, I'm stopping the tests and calling it over (for now). After 526 games, there were 265 wins to 261 losses against the original version (`rating difference: 2.64 +- 29.72 (95% confidence)`). It looks like my initial guess of parameters wasn't so good, or maybe the method itself wasn't good. Am continuing with the PUCT formula modification.

Issue #179, opened 2020-03-29
analysis engine query priority


sanderland:
Suggestion: in the json analysis program, allow a signed integer option 'priority'. Queries will be processed in order of highest priority first, tie breaking on submission order.

This is a feature suggestion that would help my program. This will help the where an SGF is submitted for analysis, and the user wants to explore variations while the program is analyzing the game. Currently you would only get feedback on these after the entire game is done analyzing, which is not ideal.

I've looked into how to add this, but would like to run it by you to get your thoughts before I put too much time into it.

* make a variant of threadsafe queue which is a priority queue
* add some counter variable which is incremented each query
* store things in the queue with key priority - epsilon*counter (or negative if it's a min-heap)
* done?

lightvector:
I think we probably just want a similar kind of thread-safe class that is built around this as the core data structure then:
https://en.cppreference.com/w/cpp/container/priority_queue

so that the ordering can be by an arbitrary type. Then we can use a tuple as the type, with the user priority being primary, and an int64 counter as the second - which would be little more "proper". :)

Otherwise, seems like a good idea. Did you want to implement it?

sanderland:
yep, that's what I meant by variant of the threadsafe queue. 
I'll give it a try, and will try to be proper. ;)

lightvector:
All done, thanks!

PR #180, opened 2020-03-30
implement priority for analysis engine

sanderland:
Implements #179 

tested and found working with https://github.com/sanderland/katrain/pull/12
after loading an sgf and rewinding, you can analyze moves near the beginning. of course there is some small delay since you need to wait for an analysis thread to free up, but it's much better than before.


lightvector:
Made some tweaks on top of this (https://github.com/lightvector/KataGo/commit/5e567ee7ed79e448c621df55cc46efe09c505af5) but this looks great. Merged and released. Thanks!

PR #186, opened 2020-04-04
Don't resign close games config parameter added

shammyx:
If the estimated lead is below resignNotCloseScore don't resign.

### DIFF
--- a/cpp/command/gtp.cpp
+++ b/cpp/command/gtp.cpp
@@ -216,7 +216,8 @@ static bool shouldResign(
   const vector<double>& recentWinLossValues,
   double lead,
   const double resignThreshold,
-  const int resignConsecTurns
+  const int resignConsecTurns,
+  const int resignNotCloseScore
 ) {
   double initialBlackAdvantageInPoints = initialBlackAdvantage(hist);
 
@@ -249,6 +250,8 @@ static bool shouldResign(
   if(pla == P_WHITE && lead > noResignationWhenWhiteScoreAbove)
     return false;
   if(resignConsecTurns > recentWinLossValues.size())
+    return false;
+  if(lead < abs(resignNotCloseScore)) //Don't resign close games
     return false;
 
   for(int i = 0; i<resignConsecTurns; i++) {
@@ -675,7 +678,7 @@ struct GTPEngine {
     Player pla,
     Logger& logger, double searchFactorWhenWinningThreshold, double searchFactorWhenWinning,
     bool cleanupBeforePass, bool ogsChatToStderr,
-    bool allowResignation, double resignThreshold, int resignConsecTurns,
+    bool allowResignation, double resignThreshold, int resignConsecTurns, int resignNotCloseScore,
     bool logSearchInfo, bool debug, bool playChosenMove,
     string& response, bool& responseIsError, bool& maybeStartPondering,
     AnalyzeArgs args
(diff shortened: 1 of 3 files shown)


shammyx:
I thought we human beings like to play a game until the end if it is kind of close, so why not adding this option here :)

lightvector:
I like the idea, I think.

* Could we maybe call it instead `resignMinScoreDifference`? Seems a little clearer.

* I think the implementation has a bug. `lead` is always from the perspective of white, in that part of the code. Therefore the logic needs to adjust the direction of the comparison based on the player.

* I would prefer to make this value commented out and defaulting to a highly negative number (so as to have no impact on the current behavior, including even the case where the lead and the winrate differ in sign). Keeping the existing default behavior is highly preferable so that KataGo works out-of-the-box for people who want to try bot vs bot matches, in which case not resigning at all is definitely a waste of computation power.

* In addition to gtp_example.cfg, this: also needs to be updated: https://github.com/lightvector/KataGo/blob/master/cpp/program/gtpconfig.cpp

I've probably spent almost as long typing this reply as it would take to just rewrite this change myself to do all the above, but whatever. :)
Does this sound fine to you, and would you like to make these fixes?

shammyx:
Yes, sure :) Tonight I'll improve the code.
- You're right, `resignMinScoreDifference` it shall be.
- Regarding the bug... By using `abs(lead)` I wanted to take care of that, but now I see I messed it up. I'll correct it.
- Commenting it out in the configs sounds good to me. Is it indeed possible that lead is differing from win rate? Anyway, a (highly) negative number as default should do fine.
Thanks for your feedback!

shammyx:
I made the changes, do you prefer me to make a second commit or should I squash them into one? I did the former but I'm happy to adjust to your style.

I just used `abs(lead)` and a default of zero now because if katago is winning - why should it surrender? So just checking `if(abs(lead) < resignMinScoreDifference)` should do.

lightvector:
Second commit is okay, thanks! I merged your change, but went ahead and made a few adjustments anyways. In particular: 

> I just used `abs(lead)` and a default of zero now because if katago is winning - why should it surrender? So just checking `if(abs(lead) < resignMinScoreDifference)` should do.

Well, it's not that simple. Whenever you write numeric code you usually want to think carefully about the corner cases. :)

Suppose that 95% of playouts in the search reach a position where white loses by 1 point, but 5% result in black blundering a huge group and white wins by 100 points in those final positions. Then, the winrate and most likely the proper result of the game should be that white loses, but the scoreLead from the search could theoretically be (0.95 * -1 + 0.05 * 100) = 4.05, as if white were "ahead" on points.

This kind of situation is very rare in practice, because that both the neural net and the search fail to resolve it with enough playouts, but it can happen. And then in such a case if using `abs(lead)`, you get non-monotone behavior. For example in the above case, if `resignMinScoreDifference=4`, white would resign because abs(4.05) > 4. But suppose you made the position _strictly worse for white_ such that white was overall 1 point worse and maybe winrate of say 4% instead of 5%. Now suddenly you would have white refuse to resign because abs(3.05) < 4 even though the position only got worse.

In general, non-monotone behavior is ugly and it's best to not add more ways for it to happen than there already are if it can be avoided by more-careful coding.

Does that make sense?

Also I made it a double instead of an int, since there's no reason why non-integer values should be rejected, but other than that, seems good.

lightvector:
(8e880d77 for the adjustments)

shammyx:
Thank you, I think I get the idea. I haven't worked on that before, so I fully trust you and value the explanation ;)
I was considering a double myself but saw no strong reason for it, so I left it as I had it initially, an int. I'm happy with either.

PR #196, opened 2020-04-12
rootInfo and option for includePolicy in analysis engine

sanderland:
Some of my finest copy-paste of code I don't fully understand, sorry if it's a bit of a mess. Not sure about the rootInfo name either.

Policy seems to give good output, in a quick test.
image


docs/Analysis_Engine.md
```diff
@@ -92,4 +198,6 @@ Explanation of fields:
       * `utilityLcb` - The LCB of the move's utility.
       * `order` - KataGo's ranking of the move. 0 is the best, 1 is the next best, and so on.
       * `pv` - The principal variation following this move. May be of variable length or even empty.
+   * `rootInfo`: A JSON dictionary with fields containing statistics for requested move itself calculated in the same way as they would be for the next moves. Current fields are: `winrate`, `scoreLead`, `scoreSelfplay`, `utility`, `lcb`, `utilityLcb`, `visits`.
```
lightvector, line 201: Should this also be optional? Or is it highly unlikely for json parsers and usage to break if this field suddenly un-optionally pops up in the analysis output?

sanderland:
It seems highly unlikely this will break parsers based on json, and given that it's basically free to include, I preferred not introducing another flag.

cpp/search/search.h
```diff
@@ -282,8 +282,10 @@ struct Search {
   void printRootEndingScoreValueBonus(std::ostream& out) const;
 
   //Get detailed analysis data, designed for lz-analyze and kata-analyze commands.
-  void getAnalysisData(std::vector<AnalysisData>& buf, int minMovesToTryToGet, bool includeWeightFactors, int maxPVDepth) const;
   void getAnalysisData(const SearchNode& node, std::vector<AnalysisData>& buf, int minMovesToTryToGet, bool includeWeightFactors, int maxPVDepth) const;
+  void getAnalysisData(std::vector<AnalysisData>& buf, int minMovesToTryToGet, bool includeWeightFactors, int maxPVDepth) const;
+  void getExtendedAnalysisData(const SearchNode& node, std::vector<AnalysisData>& buf,AnalysisData& nodeData, float (&policyProbs)[NNPos::MAX_NN_POLICY_SIZE], int minMovesToTryToGet, bool includeWeightFactors, int maxPVDepth) const;
```
lightvector, line 287: Rather than add a whole new function, why not simply have analysis.cpp call this already-existing function?
```bool getRootValues(ReportedSearchValues& values) const;  //search.h```

Also, the following three-liner should be able to get the full policy distribution from the bot., without needing to modify the code in the searchresults.cpp.
```
float policyProbs[NNPos::MAX_NN_POLICY_SIZE];
assert(search->rootNode != NULL && search->rootNode->nnOutput != nullptr); //Should always be populated after a search with at least 1 visit.
std::copy(search->rootNode->nnOutput->policyProbs, search->rootNode->nnOutput->policyProbs+NNPos::MAX_NN_POLICY_SIZE, policyProbs);
```

Maybe this is abstraction-breaking a little for the Search struct, so could warrant a function to do this, but it's also not worse than similar things that are already done elsewhere in some parts of the code, so I'd be okay with it.

sanderland:
Thanks for the pointers, will have a look tomorrow.

lightvector:
Cool, I think there shouldn't need to be any changes at all to the searchresults functions.

Yeah, I guess it's hard to see someone's usage breaking due to the new rootInfo. I guess it still might be worth explicitly mentioning in the doc that users should be resilient to the addition of new fields like that in the top-level dictionary.

sanderland:
- OK if I add utility to ReportedSearchValues?
- Dropping lcb fields as they seem to not make sense in the context of a root node.
- How do you feel about adding the pass policy value at the end of the array?

lightvector:
Adding utility would be fine. Dropping LCB-related fields also seems right. Pass should definitely be included, it's a legitimate action that a player can spend their turn to take.

sanderland:
i meant adding it at the end of the array as opposed to in a separate field like passPolicy - I definitely agree it's something to return!

lightvector:
Ah yes. Pass at the end of the array is fine. :)

sanderland:
@lightvector  should be done now, only this left as open question:
      // rootInfo["scoreMean"] = lead; // do we need 'backward compatibility' here for a new field?

lightvector:
I don't think we need scoreMean for the new field, so deleting that is fine. Thanks!

sanderland:
> I don't think we need scoreMean for the new field, so deleting that is fine. Thanks!

fixed. should be all done then.

lightvector:
Yay cool. Merged, fixed some formatting issues and made a few more clarifications to the docs, as well as and various other unrelated typo fixes in other parts of the docs too. Looks good.

PR #203, opened 2020-04-17
overrideSettings in analysis

sanderland:
started with a single setting and ended up with this.
- allows for overriding most settings in the config file with an overrideSettings field
- Unrelated, also makes the process exit nicer so you can
   ./katago analysis ... < input.in
   and it will finish all queries before exiting. I wasn't sure why it's currently killing queries so harsh when finding EOF, so happy to not have this in if there's a good reason.

Docs are todo, but thought I'd first pass this by you in case you're very opposed :)

### DIFF
--- a/cpp/command/analysis.cpp
+++ b/cpp/command/analysis.cpp
@@ -28,6 +28,7 @@ struct AnalyzeRequest {
   double rootPolicyTemperature;
   bool includeOwnership;
   bool includePolicy;
+  map<std::string,std::string> overrideSettings;
 };
 
 int MainCmds::analysis(int argc, const char* const* argv) {
@@ -70,16 +71,13 @@ int MainCmds::analysis(int argc, const char* const* argv) {
   logger.write("Analysis Engine starting...");
   logger.write(Version::getKataGoVersionForHelp());
 
+  // Load to check for unused config keys and global settings
   SearchParams params = Setup::loadSingleParams(cfg);
-  //Set a default for conservativePass that differs from matches or selfplay
-  if(!cfg.contains("conservativePass") && !cfg.contains("conservativePass0"))
-    params.conservativePass = true;
-
   const int analysisPVLen = cfg.contains("analysisPVLen") ? cfg.getInt("analysisPVLen",1,100) : 15;
-  const Player perspective = Setup::parseReportAnalysisWinrates(cfg,C_EMPTY);
   const bool assumeMultipleStartingBlackMovesAreHandicap =
     cfg.contains("assumeMultipleStartingBlackMovesAreHandicap") ? cfg.getBool("assumeMultipleStartingBlackMovesAreHandicap") : true;
   const bool preventEncore = cfg.contains("preventCleanupPhase") ? cfg.getBool("preventCleanupPhase") : true;
+  (void)Setup::parseReportAnalysisWinrates(cfg,C_EMPTY); // to stop complaints about unused keys which we will use later
 
   NNEvaluator* nnEval;
   {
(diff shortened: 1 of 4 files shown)


lightvector:
I like the idea but will need some time to read over the details, and to think about if any of the parameters changing would result in weird corners of behavior that would be bad.

I'm not sure why I made the exit process kill ongoing queries. It was definitely deliberate, but maybe for a poor reason. Is it worth making it an option or do you think having it finish stuff is just always better?

sanderland:
Possibly the reason is that if the parent process doesn't kill it's child when exiting, kata can go on to finish queries that go nowhere. Maybe a command line parameter that keeps the old behaviour is a good idea.

lightvector:
Looks good. There were a few subtle things left to address that are unfortunate complications. I handled them in this commit. Feel free take a look at the diffs and let me know what you think! 

6ed9bed4

Details:
* A few of the search parameters are not changeable after startup, so I added a check for that.

* I made it optional to keep the old behavior.

* Also this old assert actually I discovered during testing could fire for the old behavior. Pre-existing bug for the old behavior, not related to this latest pull request. 
`assert(search->rootNode != NULL && search->rootNode->nnOutput != nullptr);`

So I fixed that, which required some various work on the threading of asyncbot to behave better, which were probably worth making anyways - the cleanup logic for asyncbot was a bit loose before.

* I got rid of your sleep loops. Those aren't needed. Just changing the status on the queue is enough.

lightvector:
And then in this commit:
e9f1eecb

I moved the parsing of the overrided config into the main parsing loop, rather than in the analysis loop. This allows cleaning up all the other ad-hoc parameters out of the AnalyzeRequest, which is overall a much cleaner design than even I had originally. This also fixes a subtle bug in the original implementation, where `rootPolicyTemperature` would override `rootPolicyTemperatureEarly` causing silent ignoring of one of them if you tried to set different values for them in the config. Yay.

lightvector:
Let me know if you encounter any bugs, or happen to notice anything that looks like a messup by me in those diffs.

lightvector:
And some docs and safeguard against typos here:
592d06f5

sanderland:
maxTime override works
PDA override does not generate the sawtooth results it used to, something definitely seems wrong here.

```
Sending query QUERY:2: {'rules': 'japanese', 'priority': 1, 'analyzeTurns': [1], 'maxVisits': 5000, 'komi': 6.5, 'boardXSize': 19, 'boardYSize': 19, 'includeOwnership': False, 'includePolicy': True, 'moves': [['B', 'Q16']], 'overrideSettings': {'maxTime': 1.0, 'playoutDoublingAdvantage': 3.0, 'playoutDoublingAdvantagePla': 'WHITE'}, 'id': 'QUERY:2'}

,"rootInfo":{"scoreLead":-20.412331547019885,"scoreSelfplay":-23.652722253045283,"scoreStdev":22.660806192157658,"utility":-1.021435270387918,"visits":49,"winrate":0.021515084943533935},"turnNumber":1}\n'
,"moveInfos":[{"lcb":0.9646034748896277,"move":"R4","order":0,"prior":0.10426804423332214,"pv":["R4","D4","C16","P3"],"scoreLead":20.642114045331674,"scoreMean":20.642114045331674

```
where I would expect -20/-20 for both root and move

and similar results (i.e. rootinfo score -20, moveinfo +20 with a query using
{'maxTime': 1.0, 'playoutDoublingAdvantage': 3.0, 'playoutDoublingAdvantagePla': 'BLACK'}, , where I would expect scores of +20 for black.
all queries use reportAnalysisWinratesAs = BLACK

sanderland:
image
i think it's dropping reportAnalysisWinratesAs somehow

lightvector:
Thanks for catching this.

PR #209, opened 2020-04-26
Fix perspective in  analysis engine

sanderland:
It is now always C_EMPTY due to 

- not updating the new query from rbase
- ~~not setting a default from the global config in the loadParams override~~


cpp/command/analysis.cpp
```diff
@@ -704,6 +704,7 @@ int MainCmds::analysis(int argc, const char* const* argv) {
         newRequest->hist = hist;
         newRequest->nextPla = nextPla;
         newRequest->params = rbase.params;
+        newRequest->perspective = rbase.perspective;
```
lightvector, line 707: This is the only line that needs to be fixed, right?

lightvector: (that is, the other changes are unnecessary, we just need only this single line as the entire commit to fix this bug, right?)

sanderland: you're right, this was the last line I changed before thinking it was the other ones, but the cfg takes care of that. fixed.

lightvector:
Accepted, but squashed into one commit as 3016348e.
Thanks!

Issue #225, opened 2020-05-12
Linking error on Linux/CUDA version 


cleinias:
When trying to build the CUDA version on Archlinux, I get a linking error in the early test phase. I can compile the Nvidia CUDA samples successfully, so it does not seem to be a CUDA  installation issue.   The linker seems unable to find the dl library, as far as I can tell. 

~~~~
Building 'katago' executable for GTP engine and other tools
 -DUSE_BACKEND=CUDA, using CUDA backend
 The CUDA compiler identification is NVIDIA 10.2.89
 Check for working CUDA compiler: /opt/cuda/bin/nvcc
 Check for working CUDA compiler: /opt/cuda/bin/nvcc - broken
 CMake Error at /usr/share/cmake-3.17/Modules/CMakeTestCUDACompiler.cmake:46 (message):
   The CUDA compiler

     "/opt/cuda/bin/nvcc"

   is not able to compile a simple test program.

   It fails with the following output:

     Change Dir: /home/stefano/builds/KataGo/cpp/CMakeFiles/CMakeTmp

     Run Build Command(s):/usr/bin/make cmTC_18138/fast && /usr/bin/make  -f CMakeFiles/cmTC_18138.dir/build.make
 CMakeFiles/cmTC_18138.dir/build
     make[1]: Entering directory '/home/stefano/builds/KataGo/cpp/CMakeFiles/CMakeTmp'
     Building CUDA object CMakeFiles/cmTC_18138.dir/main.cu.o
     /opt/cuda/bin/nvcc -forward-unknown-to-host-compiler   -cudart shared    -x cu -c
 /home/stefano/builds/KataGo/cpp/CMakeFiles/CMakeTmp/main.cu -o CMakeFiles/cmTC_18138.dir/main.cu.o
     Linking CUDA executable cmTC_18138
     /usr/bin/cmake -E cmake_link_script CMakeFiles/cmTC_18138.dir/link.txt --verbose=1
     /opt/cuda/bin/g++   CMakeFiles/cmTC_18138.dir/main.cu.o -o cmTC_18138  -lcudadevrt -lcudart_static
 -L"/opt/cuda/targets/x86_64-linux/lib/stubs" -L"/opt/cuda/targets/x86_64-linux/lib" -lcudadevrt -lcudart
     /usr/bin/ld: /opt/cuda/targets/x86_64-linux/lib/libcudart_static.a(libcudart_static.a.o): undefined reference to symbol
[...]

PR #227, opened 2020-05-13
Workaround CMake 3.17.0 (Bug? Errata? API Change?) with handling CUDA…

OmnipotentEntity:
… compiler testing

See issue #225 for details, and thanks @cleinias for the patch

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -38,12 +38,15 @@ message("Building 'katago' executable for GTP engine and other tools")
 if(USE_BACKEND STREQUAL "CUDA")
   message("-DUSE_BACKEND=CUDA, using CUDA backend")
 
-  # Ensure dynamic cuda linking
-  set(CMAKE_CUDA_FLAGS "" CACHE STRING "")
-  if(CMAKE_CUDA_FLAGS)
-    list(REMOVE_ITEM CMAKE_CUDA_FLAGS "-cudart static")
+  
+  # Ensure dynamic cuda linking (Versions prior to 3.17)
+  if (${CMAKE_VERSION} VERSION_LESS "3.17")
+    set(CMAKE_CUDA_FLAGS "" CACHE STRING "")
+    if(CMAKE_CUDA_FLAGS)
+      list(REMOVE_ITEM CMAKE_CUDA_FLAGS "-cudart static")
+    endif()
+    string(APPEND CMAKE_CUDA_FLAGS "-cudart shared")
   endif()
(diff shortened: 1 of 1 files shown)


OmnipotentEntity:
I should note that this PR has been tested and seems to work on older versions of CMake as well.

lightvector:
Thanks. I'll try it myself soon.

OmnipotentEntity:
To be clear, I think this is a bug in CMake, and I'm not necessarily expecting this patch to be accepted.  Rather, I think eventually CMake will fix this issue, and this patch will not be necessary.

I'm submitting this patch for review specifically because NixOS has a policy that any patches made against upstream should be submitted to the maintainers, and not kept out of band.

I expect the next time a katago release comes around this patch won't actually be needed.

OmnipotentEntity:
CMake tracking issue: https://gitlab.kitware.com/cmake/cmake/-/issues/20708

OmnipotentEntity:
Per CMake devs, this is a better workaround, and probably the proper way of making CUDA shared.

See [this comment in the tracking issue for a rundown.](https://github.com/lightvector/KataGo/issues/225#issuecomment-630246169)

OmnipotentEntity:
Fixed this specificity of the version.  This should now work for everyone.

suyoung-park:
Dear sir


I want to contact you. But I don't know how. So I write here
 

We are one of the broadcasting channel in Korea, specialized in Baduk(=GO), 

which is called ‘KBaduk’.

 

We want to provide winning-rate graphs to our viewers when we broadcast, 

using your ‘KataGo (Copyright 2019 David J Wu ("lightvector")’ system.

Of course, we will notate <Sources from KataGo> at the bottom of the graphs.

 

You've told me when using KataGo, it was mentioned that “Permission is granted including without limitation 

the right to use, copy, modify, merge, publish, distribute,sublicense, and/or sell copies of the Software”, 

 

However KBaduk want to use KataGo with your permission. 

Is it able to use KataGo and notate that the source is from KataGo when we broadcast? 

I am looking forward to your reply. 

If you can't reply here, please send an email to l psypassion@gmail.com 

Take care of yourself.


Yours sincerely.

lightvector:
@suyoung-park - Sounds cool. I'm replying to you privately. In the future, if you want to post a new question, please make a new separate issue (Issues -> New Issue), rather than posting in an issue or pull request that is unrelated to your topic.

lightvector:
Maybe it's not needed now, and mabe this is a bit late, but I merged it anyways in case anyone in the future lands on one of these cmake versions. I think Github didn't auto detect it though, due me also doing a manual rebase, so I'm going ahead to just close this. Thanks!

PR #295, opened 2020-08-11
edited cudabackend.cpp to make it run on cuDNN8,

Harald-Han:
which supports CUDA 11.0 and 10.2;
it is no longer compatible with cuDNN7, though

### DIFF
--- a/cpp/neuralnet/cudabackend.cpp
+++ b/cpp/neuralnet/cudabackend.cpp
@@ -1,3 +1,4 @@
+// modified by Harald Han to make it compatible with CUDNN 8 but it no longer works with CUDNN 7
 #ifdef USE_CUDA_BACKEND
 #include "../neuralnet/cudaerrorcheck.h"
 #include "../neuralnet/cudaincludes.h"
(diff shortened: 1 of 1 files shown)


petgo3:
Hi there, I'm not too sure this is relevant, but just to doublecheck it:
I installed cuda 11.0 and used this version on windows 10 using cMake 3.15 and vs 2019.
I got "Unsupported gpu architecture 'compute_30' " which was generated in "CMakeLists.txt". 
I solved this by deleting the compute_30 part of 

 set(CMAKE_CUDA_FLAGS
    "-gencode arch=compute_37,code=sm_37 -gencode arch=compute_53,code=sm_53 -gencode arch=compute_53,code=compute_53 -gencode arch=compute_70,code=sm_70 -gencode arch=compute_70,code=compute_70"
    )

Now it works like a charm :-)

Harald-Han:
Yes, CUDA 11 doesn't support older GPU architectures any more so that CMakeLists.txt should be edited accordingly.

lightvector:
Big question: what should we do for users that haven't or aren't able to upgrade to CUDA 11 due to their OS or due to other dependencies?

Harald-Han:
I think that checking the version of the cuDNN library at compile-time and having the appropriate part of the code compiled is the only solution to support both cuDNN 7 and cuDNN 8.

lp200:
V100 seems to be a little slower.
But maybe A100 will need cudnn8

lightvector:
Closing this since I believe is it made redundant by https://github.com/lightvector/KataGo/pull/317

lightvector:
Thanks!

PR #299, opened 2020-08-18
Fix MSVC compile error

### LINKED ISSUE #296: Error when compile eigen backend on MSVC
I can complie kaorahi's borrow_tensorflow branch with newest eigen3 or 1.50 release tag with eigen3.3.7 release
but compile current master branch,i get this error
微信截图_20200813145908
I have tired eigen3.3.7 and newest eigen3,both get this error.
	C2664	“void Eigen::Tensor<float,2,0,Eigen::DenseIndex>::resize(void)”: Cannot convert “const std::array<int,2>”to “__int64”	katago	C:\Program Files (x86)\Eigen3.7\include\eigen3\unsupported\Eigen\CXX11\src\Tensor\Tensor.h	416

sanderland:
Fixes #296


lightvector:
I *think* this should be handled now.  The right type appears to be Eigen::Index, and also on *my* version of MSVC I actually get more compile errors besides this, relating to initializing the arrays via `( { initializer list elements } )` instead of `= { initializer list elements }`.

If the new tip of master works with your MSVC version too, then I think everything is good now.

sanderland:
have merged it into my version and will test later, but assume it's fine.

PR #301, opened 2020-08-19
Fix cout typos in opencltuner.cpp

sanderland:


lightvector:
Excellent, thanks for catching this!

PR #317, opened 2020-09-18
Add support for CUDA 11 and cuDNN8 (without breaking CUDA 10 / cuDNN7)

anoek:
This patch builds on #295 by @Harald-Han by adding some macro tests to use the appropriate code depending on whether we are compiling against cuDNN7 or cuDNN8. I've also updated the CMakeLists.txt file to use the appropriate gencode settings depending on whether we are using CUDA 10.x or CUDA 11.x.

This should be fully compatible with all existing builds out there, and of course now adds compatibility with systems running with the newer CUDA versions and should add support for Ampere devices with the new gencode settings.

I've tested the build on both CUDA 10/cuDNN 7 and CUDA 11/cuDNN 8 systems.

I have not personally validated the correctness of the cuDNN8 port, but it seemed pretty straight forward and looked good from my naive perspective, however I don't have enough experience with cuDNN myself to give a confident stamp of approval.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -74,9 +74,66 @@ if(USE_BACKEND STREQUAL "CUDA")
   endif()
 
   set(NEURALNET_BACKEND_SOURCES neuralnet/cudabackend.cpp neuralnet/cudahelpers.cu)
-  set(CMAKE_CUDA_FLAGS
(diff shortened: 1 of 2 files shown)


lightvector:
Excellent, I will take a look at this soon. Thanks!

lightvector:
I'm not currently set up to actually the CUDNN8, but I'll assume this is good and everything works fine for me back on CUDNN7, so yay.

I also pushed some changes to hopefully not break people who were working CUDA even older than 10.2, since I think there are still some people building from older versions and having it work: 8bdb2dcd

PR #322, opened 2020-09-27
Return ownership for all candidate moves

sanderland:
This includes the ownership map in all 'moveInfos' objects for the analysis engine.

Probably needs some cleaning up and could be optimized, but opening this PR to see if you're interested in including this.
I'm using it to experiment with 'simple moves' AIs on the front-end side.

### DIFF
--- a/cpp/command/analysis.cpp
+++ b/cpp/command/analysis.cpp
@@ -28,12 +28,33 @@ struct AnalyzeRequest {
   Player perspective;
   int analysisPVLen;
   bool includeOwnership;
+  bool includeMovesOwnership;
   bool includePolicy;
   bool includePVVisits;
 
   vector<int> avoidMoveUntilByLocBlack;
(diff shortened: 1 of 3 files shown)


lightvector:
Yeah, seems plausible. Yeah, maybe would be good to do in a way that factors out some of the logic into a common function and worth thinking about optimization and avoiding duplicating work with main ownership and the amount of data sent back. But I could certainly see adding this option.

sanderland:
Cleaned it up a bit, factored out formatting function and such, and introduced a documented option for doing this.
Had a look at if the root+children ownership could easily be retrieved in one go, and found the answer to be no, probably not worth introducing bugs over premature optimization.

lightvector:
Merged, along with a few additional cleanups in 656414c8

Thanks!

lightvector:
Including, if it matters, one slight change to the semantics - the per-move ownership option no longer implies the root ownership option. They're just independent options now.

PR #335, opened 2020-10-25
Remove Boost library dependency

isty2e:
Currently the only thing KataGo is utilizing from boost is the `filesystem` library. I thought having a dependency on such a heavyweight library is not the greatest thing in the world. Therefore:

- [Filesystem library](https://github.com/gulrak/filesystem) was used instead, which is a header-only implementation of C++17 `std::filesystem` for C++11, C++14 and C++17.
- `filesystem.hpp` was added in `/external/filesystem-1.3.6/include/ghc/`.
- `<boost/filesystem.hpp>` was replaced to `"../external/filesystem-1.3.6/include/ghc/filesystem.hpp"`.
- `boost::filesystem` was replaced to `ghc::filesystem`.

I tested it on a linux system and it worked fine, but haven't really tested on a Windows system. Also this change raises some warning when making due to switches in `filesystem.hpp` missing default cases, so one might want to remove `-Wswitch-default` flag, but I'm not sure if this is adequate.

Edit: Tested on a Windows system.


l1t1:
how to get the source tar ball with this commit ?

isty2e:
> how to get the source tar ball with this commit ?

You can just `git clone`.

lightvector:
Merged! Yay!

PR #345, opened 2020-11-11
Fix build on aarch64 and aarch32 systems

anoek:
The `CMAKE_SYSTEM_PROCESSOR` cmake variable takes on a few different values for different arm systems, so this patch just adds those values in.

For anyone curious, this was the only modification needed to get KataGo to build on NVidia's Jetson Nano 2G board, which detects as `aarch64`. The device benchmarks at 11.32 visits/s :)


lightvector:
Thanks!

PR #362, opened 2020-12-04
Update modelconfigs.py

petgo3:
comma typos


lightvector:
Thanks for the quick notice, this is fixed by https://github.com/lightvector/KataGo/commit/dc2ef639046a439e79559dcacc0df4e939d29e9c. I want to keep the trailing commas in a few places since that makes the code more uniform and reduces future missing commas. :)

PR #375, opened 2020-12-27
Move whiteNoResultProb assignment to match rest of codebase

roy7:
This assignment seemed to be out of order vs all other parts of the codebase.

Totally unimportant nit I stumbled across.


lightvector:
Yes, good catch, but also keeping it where it is means we don't rerandomize all of the tests... hmm....

roy7:
Can leave it if you like, it's purely cosmetic.

PR #377, opened 2020-12-27
Rework how symmetry is used to support random selection without replacement.

roy7:
Should the shuffle pass the 'rand' random number generator? I tried doing it simply, but it wouldn't compile.

I don't really understand the getHash code section. But if we add nnInputParams.symmetry to the calculation of the hash, we don't need to turn cache off any more when pulling more than 1 symmetry to average. I put a comment about it within the PR.

Intent of this is to support averaging multiple symmetry together, without repeating the same symmetry more than once.


roy7:
Since we're only adding an int per result buffer (batch size) and the MiscNNInputParams, there should be minimal impact on performance or use of memory.

cpp/search/search.cpp
```diff
@@ -2003,7 +2003,12 @@ void Search::initNodeNNOutput(
   }
   if(isRoot && searchParams.rootNumSymmetriesToSample > 1) {
     vector<shared_ptr<NNOutput>> ptrs;
+    std::array<int, NNInputs::NUM_SYMMETRY_COMBINATIONS> symmetryIndexes {0,1,2,3,4,5,6,7};
+    // Should this pass rand as third param? Wouldn't compile.
```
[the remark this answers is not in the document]
roy7: Is it worth adding a loop to support NUM_SYMMETRY_COMBINATIONS with value other than 8? Or is hard coding {0,1,2,3,4,5,6,7} acceptable?

roy7:
Is it worth adding a loop to support NUM_SYMMETRY_COMBINATIONS with value other than 8? Or is hard coding {0,1,2,3,4,5,6,7} acceptable?

cpp/search/search.cpp
```diff
@@ -2003,7 +2003,12 @@ void Search::initNodeNNOutput(
...
+    if (searchParams.rootNumSymmetriesToSample < 8) std::random_shuffle(symmetryIndexes.begin(), symmetryIndexes.end());
     for(int i = 0; i<searchParams.rootNumSymmetriesToSample; i++) {
+      nnInputParams.symmetry = symmetryIndexes[i];
+      // If we add symmetry to the nnHash calculation, no need to skip caches when i > 0?
```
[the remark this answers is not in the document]
roy7: As far as I can tell from staring at the code a bit, hash is calculated at the evaluate() level and isn't searched for transpositions. The symmetry selection is done within getOutput() after the cache is already checked and hash calculated. So the hash doesn't change based on the actual symmetry used or board transposition sent to the gpu for eval.

As such, are transpositions not currently using cached values anyway? Or am I missing something obvious? My expectation was that there would be a search of all 8 transposition hashes against the cache, someplace. So opening moves ABC and ACB in search would end up sharing that eval.

cpp/search/search.cpp
```diff
@@ -2003,7 +2003,12 @@ void Search::initNodeNNOutput(
   }
   if(isRoot && searchParams.rootNumSymmetriesToSample > 1) {
     vector<shared_ptr<NNOutput>> ptrs;
+    std::array<int, NNInputs::NUM_SYMMETRY_COMBINATIONS> symmetryIndexes {0,1,2,3,4,5,6,7};
+    // Should this pass rand as third param? Wouldn't compile.
+    if (searchParams.rootNumSymmetriesToSample < 8) std::random_shuffle(symmetryIndexes.begin(), symmetryIndexes.end());
```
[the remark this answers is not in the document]
roy7: Corrected per your other suggestion.

cpp/neuralnet/nninputs.h
```diff
@@ -31,6 +31,7 @@ struct MiscNNInputParams {
   double playoutDoublingAdvantage = 0.0;
   float nnPolicyTemperature = 1.0f;
   bool avoidMYTDaggerHack = false;
+  int symmetry = -1;
```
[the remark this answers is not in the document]
roy7: Comments added

cpp/search/search.cpp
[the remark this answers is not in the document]
roy7: Ok right, I'm using the wrong term and katago doesn't try to detect symmetry when using the cache. So I don't need to worry about any of that.

roy7:
I've create a nodeNumSymmetriesToSample branch as well, which is built off this one. I'll PR it when/if these changes are accepted. :)

cpp/search/search.cpp
```diff
@@ -2003,7 +2003,10 @@ void Search::initNodeNNOutput(
   }
   if(isRoot && searchParams.rootNumSymmetriesToSample > 1) {
     vector<shared_ptr<NNOutput>> ptrs;
+    std::array<int, NNInputs::NUM_SYMMETRY_COMBINATIONS> symmetryIndexes {0,1,2,3,4,5,6,7};
```
sanderland, line 2006: this feels strange to me, with NUM_SYMMETRY_COMBINATIONS being a variable, but the elements being hardcoded. std::iota ?

roy7: Yeah I can change this to populate it with a loop if you prefer.

lightvector: I think it's fine, as long as it has the property of failing to compile if NNInputs::NUM_SYMMETRY_COMBINATIONS were to ever somehow change.

lightvector: (i.e. the critical thing is that we are never accidentally in a state where the two become inconsistent and compile without knowing it - either they should automatically be consistent, or else it should be the case that they aren't, but the inconsistent state fails to compile so that it cannot be accidentally missed).

roy7: I like Sander's idea to populate it with std::iota. Is it just this?

std::iota(symmetryIndexes.begin(), symmetryIndexes.end(), 0);

roy7: Change made.

roy7:
Ok, I've also added a couple defines to remove the reliable on magic -1 numbers related to symmetry. Should make the code more clear to people viewing it.

cpp/neuralnet/nninputs.h
```diff
@@ -10,6 +10,9 @@
 #include "../game/boardhistory.h"
 #include "../game/rules.h"
 
+#define SYMMETRY_NOTSPECIFIED -1
```
lightvector, line 13: Can we move these two constants into the NNInputs namespace, and make them instead constexpr int?

cpp/search/search.cpp
```diff
@@ -2003,7 +2004,11 @@ void Search::initNodeNNOutput(
   }
   if(isRoot && searchParams.rootNumSymmetriesToSample > 1) {
     vector<shared_ptr<NNOutput>> ptrs;
+    std::array<int, NNInputs::NUM_SYMMETRY_COMBINATIONS> symmetryIndexes;
+    std::iota(symmetryIndexes.begin(), symmetryIndexes.end(), 0);
     for(int i = 0; i<searchParams.rootNumSymmetriesToSample; i++) {
```
lightvector, line 2009: Before, rootNumSymmetriesToSample > 8 would have been valid and meaningful. Now, it will give undefined behavior. This function should probably explictly check for that and fail, either via throwing an exception (KataGo code just uses a custom "StringError" exception for pretty much everything) or via an assert.

And additionally, the bounds code in setup.cpp should be tightened to reject values larger than 8 now.

Is it worth having a parameter to toggle with and without replacementness, to preserve old behavior? I'm not sure. I guess probably nobody was using values larger than 8 here in practice.

cpp/neuralnet/nninputs.h
```diff
@@ -31,6 +34,8 @@ struct MiscNNInputParams {
   double playoutDoublingAdvantage = 0.0;
   float nnPolicyTemperature = 1.0f;
   bool avoidMYTDaggerHack = false;
+  // If no symmetry is specified, it will use default or random based on config.
+  int symmetry = SYMMETRY_NOTSPECIFIED;
```
lightvector, line 38: Should mention that this value may be ignored even when specified, if there is a cache hit, if not also specifying skipCache?

cpp/search/search.cpp
```diff
@@ -2003,7 +2004,11 @@ void Search::initNodeNNOutput(
...
+      std::swap(symmetryIndexes[i], symmetryIndexes[thread.rand.nextInt(i,NNInputs::NUM_SYMMETRY_COMBINATIONS-1)]);
+      nnInputParams.symmetry = symmetryIndexes[i];
       bool skipCacheThisIteration = skipCache || i > 0; //Skip cache on subsequent iterations to get new random draws for orientation
```
lightvector, line 2012: Seems like this is going to be buggy because if the cache already has this position, on iteration 0 we won't get the symmetry we asked for, right?

It used to be okay to trust the cache for iteration 0 because we were sampling without replacement, the cached value was as good as any other arbitrary draw for the first random symmetry. But now that we have a specific array we have shuffled and that we are relying on getting an exact specific symmetry, we can no longer use the cache for *any* of the iterations, right?

Seems like this is going to be a very slight extra performance/latency cost on the root node if we can't use the cache even on iteration 0 (because the root node also blocks any possible multithreading to search the rest of the tree too). It makes this feature somewhat more expensive than before. But maybe it's not too big a deal.

lightvector (changes requested):
I like the idea a lot, and having named constants is much better.

Left some more comments and questions, including a missing safety check and also what seems to me like a potential bug in the current logic.

cpp/search/search.cpp
roy7: I guess I made the changes in the other later PR instead of this one, but per @sanderland feedback I did change the config file processing to limit the maximum to NNInputs::NUM_SYMMETRY_COMBINATIONS. So it wouldn't be possible to specify one that was too high at startup. (https://github.com/roy7/KataGo/pull/1/files)

I doubt the old behavior is worth keeping around. Anyone using a # more than 1 currently, I'd assume without replacement is strictly better. And I just can't imagine someone using a number over 8 and intentionally wanting replacement instead of without.

cpp/neuralnet/nninputs.h
roy7: Done

cpp/search/search.cpp
roy7: setup.cpp updated

cpp/neuralnet/nninputs.h
roy7: Note added

cpp/search/search.cpp
roy7: Cache always skipped in this part of the code now.

cpp/search/search.cpp
roy7: With setup updated to cap at NUM_SYMMETRY_COMBINATIONS I assume that should guarantee good behavior, but I can add an assert if you like.

roy7:
I'm not actively running the code since I don't operate Katago bots on OGS, and I'm not sure how a test would have found it. The original "if i>0" was a bit of a shortcut anyway though, relying on the fact we don't know what symmetry is in the cache since we cache at a higher level than raw input going into the neural network. With replacement, we didn't care, but it was a bit of a performance hack than an intentional design decision?

Maybe someone like @dorus would be willing to import the PR for a bit and run it live on OGS to make sure no expected behaviors/problems get reported.

lightvector:
Maybe there should be a test that setting the number of symmetries to 8 should be deterministic for the final result (except for tiny floating point errors in the last decimal place possibly?). That's at least one test I can think of.

Another would be to run a bunch of 2s, and verify that in all cases that they are one of the 28 possible averages of the 8 possible. I could take the time to write that test, unless you're up to try it.

lightvector:
(I think both of those tests would have had a good chance to catch that particular bug, at least)

roy7:
I've never written a test before, but can try to learn if you like. :) It makes sense to check that 8 is deterministic since there's no reason to randomly shuffle the order when doing all of them.

lightvector:
So there are actually some existing tests that do root symmetry averaging and I'm trying to reconcile those tests with this. I think there is another subtle semantic change that is causing problems - which is that meaning of doRandomize has changed. Previously setting it to false locked the symmetry choice to a single value, regardless of things like rootNumSymmetriesToSample. Now, with this change, it doesn't.

There are still various comments in the code and tests that are written to that effect that are now wrong. This is making things a bit messy. I'm not confident I'm going to be able to merge this before the next release.

lightvector:
Or, I could revert that bit. It seems sort of orthogonal to selecting without replacement. Anyways, the fact that there's also this subtle semantic change and the tests are all being rerandomized (due to a different number of calls to the seeded rand) at the same time also makes me also nervous. Maybe it would have been worth doing the purely structural refactors entirely separately from the semantic changes, so it would be more clear that those refactors aren't breaking anything.

lightvector:
Hmmm, actually maybe there's a way to do this that works, making use of pre-arranging and regenerating tests to not depend on the state of the RNG seed before merging with this branch. :)

roy7:
Well I'm glad the tests are complex enough to catch all sorts of things even if it means my changes will take longer to possibly to get added. There's no urgency. I can also start experimenting in my own fork once I decide to dive in. All of this (and my future followup nodeNum PR) are just to lay the groundwork tools I'd need. They seem reasonable to add to the main code and won't harm performance when not used, in theory.

roy7:
I wonder, maybe reset the seed in each section of the tests? Either to the same seed, or a section specific seed. So if something changes in one area of the code that changes the number of rand() calls, it won't impact any other existing tests in other sections of the code?

lightvector (approved):

lightvector:
Merged, along with various changes to tests, and some more cleanups, for example: a6a79c01 and 5011f1d2 and 732f29b6 and c3a99a20. If you want to take a peek at how the tests work, feel free to take a look at these.

lightvector:
Anyways, I think this is good. Tests seem fine, happy with the code change and improved symmetry logic. :)

roy7:
Yay thanks! We'll see if the next PR is as clean. ;)

PR #380, opened 2020-12-28
distributed client console & json logging

sanderland:
Adds
* Console logging as an alternative to watchfile - this was an intermediate step and I'm happy to remove it, but it could also be useful.
* JSON logging of moves, in an analysis-engine like format. A bit hardcoded on config still, but works well enough.
* logOwnership parameter to playGame seems a tad messy, but not sure where else it belongs (searchParams?)

Following our discussion, i just disabled encore/cleanup output for the json side.


lightvector (commented):
You might be interested in the `demoplay` command in misc.cpp, which went even further and could generate logs to be played back in real-time about how the visits distribution and ownership map evolved *within* each ongoing search. It used to feed to a python visualizer that printed it in a Lizzielike format, but more colorful, with ongoing graphs and stats, and a special board dedicated to the ownership heatmap so that it could be displayed much more prominently.

The python side for demoplay though unfortunately was never made public, and is not part of the codebase that I have access to any more (this was in part related to earlier sponsorship stuff).

But anyways you can see the KataGo side of it and how it works. It's done through specifying a periodic callback:
```
    std::function<void(const Search*)> callback = [&baseHist,&recentWinLossValues,&recentScores,&recentScoreStdevs](const Search* search) {
      writeLine(search,baseHist,recentWinLossValues,recentScores,recentScoreStdevs);
    };
```

cpp/game/rules.h
```diff
@@ -4,6 +4,9 @@
 #include "../core/global.h"
 #include "../core/hash.h"
 
+#include "../external/nlohmann_json/json.hpp"
+using json = nlohmann::json;
```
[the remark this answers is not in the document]
sanderland: I could 
* json load the string that rules generates (a little sad)
* just have the function in contribute (very sad)

cpp/command/contribute.cpp
```diff
@@ -171,15 +176,154 @@ static void runAndUploadSingleGame(
       (void)scoreStdevHist;
       (void)search;
       out << "\n";
-      (*outputEachMove) << out.str() << std::flush;
+      if(outputEachMove != nullptr)
+        (*outputEachMove) << out.str() << std::flush;
+      if(watchOngoingGameInConsole)
+        std::cout << out.str() << std::flush; 
     }
+    if(logGamesAsJson and hist.encorePhase == 0) { // If anyone wants to support encorePhase > 0 note passForKo is a thing
```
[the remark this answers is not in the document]
sanderland: Yes it does

cpp/configs/contribute_example.cfg
```diff
@@ -16,6 +16,14 @@ maxSimultaneousGames = Fill in a number here
 watchOngoingGameInFile = false
 watchOngoingGameInFileName = watchgame.txt
 
+# Set to true if you want one of KataGo's games to be streamed to the console, so you can watch.
+watchOngoingGameInConsole = false
```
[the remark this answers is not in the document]
sanderland: Yes it does, the log is completely swamped by this. OTOH it's quick and easy and doesn't leave huge log files.

sanderland:
> You might be interested in the `demoplay` command in misc.cpp, which went even further and could generate logs to be played back in real-time about how the visits distribution and ownership map evolved _within_ each ongoing search. It used to feed to a python visualizer that printed it in a Lizzielike format

That's definitely pretty cool, but i imagine the self-play is fast enough that this will be a lot of work for little interest for the average user (who is essentially getting a screensaver in return for contributing)

cpp/command/contribute.cpp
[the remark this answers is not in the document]
sanderland: done

cpp/search/search.h
```diff
@@ -18,6 +18,8 @@
 #include "../search/searchprint.h"
 #include "../search/timecontrols.h"
 
+#include "../external/nlohmann_json/json.hpp"
```
sanderland, line 21: This probably has the same problem as including this in rules, but it's not clear where else these functions belong. A recommendation for a home would be helpful.

lightvector:
> > You might be interested in the `demoplay` command in misc.cpp, which went even further and could generate logs to be played back in real-time about how the visits distribution and ownership map evolved _within_ each ongoing search. It used to feed to a python visualizer that printed it in a Lizzielike format
> 
> That's definitely pretty cool, but i imagine the self-play is fast enough that this will be a lot of work for little interest for the average user (who is essentially getting a screensaver in return for contributing)

Oh no, being fast enough is not the point. The point is that seeing a smoother transition, and being able to make out distinctly the transitions KataGo is changing its mind or reconsidering a variation or what the result of something will be can actually be kinda cool if visualized. 

See these two for example for just the ownership alone, not including showing how the policy is changing its mind, which can be really cool when smooth and you can see the moment of shift:

https://user-images.githubusercontent.com/11942395/103439460-2d1a9480-4c0b-11eb-9b4d-ed54c91cd9b3.mp4
https://user-images.githubusercontent.com/11942395/103439469-33107580-4c0b-11eb-8dfe-62deb3175adf.mp4

(note that one would actually probably want to use one of those those perceptually-uniform color scales instead of the actual one there, and probably reduce the saturation slightly so that the scales anchor at something closer to just black/white - but I've found having the ownership visualization twist partly through color-space as well instead of being only black and white can both look cool and also make differences easier to see, because having hue change *in addition* to brightness means makes changes easier to discriminate).

sanderland:
> Oh no, being fast enough is not the point. The point is that seeing a smoother transition, and being able to make out distinctly the transitions KataGo is changing its mind or reconsidering a variation or what the result of something will be can actually be kinda cool if visualized.

Ah so, I can definitely see report during search being a thing here later on, but maybe a bit too far at this point.

cpp/configs/contribute_example.cfg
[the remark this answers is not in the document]
lightvector: Let's remove it, if you only need the json part. Thanks.

sanderland:
json output back in rules and used that function to clean up the repetitiveness of the old ones a bit as well. Should be ready now.

cpp/search/searchresults.cpp
```diff
@@ -1311,3 +1310,144 @@ double Search::getAverageTreeOwnershipHelper(vector<double>& accum, int64_t minV
...
+  vector<AnalysisData> buf;
+  getAnalysisData(buf, minMoves, false, analysisPVLen);
+
+  // Stats for all the individual moves
+  json moveInfos = json::array();
+  for(int i = 0; i < buf.size(); i++) {
+    const AnalysisData& data = buf[i];
+    double winrate = 0.5 * (1.0 + data.winLossValue);
+    double utility = data.utility;
+    double lcb = 0.0; //    PlayUtils::getHackedLCBForWinrate(this, data, pla);
```
lightvector, line 1354: What is the reason for commenting this out?

cpp/configs/contribute_example.cfg
```diff
@@ -16,6 +16,11 @@ maxSimultaneousGames = Fill in a number here
 watchOngoingGameInFile = false
 watchOngoingGameInFileName = watchgame.txt
 
+# Set to true if you want one of KataGo's games to be output as JSON, so you can programatically process/visualize them
```
lightvector, line 19: Do you mind if I delete all of this in the example cfg? Seems not so useful for a common user to be configuring the option to dump an enormous amount of data or not to stdout. KaTrain can pass it using --override-config where needed when launching the program.

cpp/command/analysis.cpp
```diff
@@ -254,130 +235,19 @@ int MainCmds::analysis(int argc, const char* const* argv) {
...
-    }
-    ret["moveInfos"] = moveInfos;
-
-    // Stats for root position
-    {
-      ReportedSearchValues rootVals;
-      bool suc = search->getRootValues(rootVals);
-      if(!suc)
-        return false;
-      Player rootPla = getOpp(request->nextPla);
```
lightvector, line 330: Flipping the player here seems like a bug that existed in the old code as well.

cpp/search/searchresults.cpp
```diff
@@ -1311,3 +1310,144 @@ double Search::getAverageTreeOwnershipHelper(vector<double>& accum, int64_t minV
...
+  }
+  ret["moveInfos"] = moveInfos;
+
+  // Stats for root position
+  {
+    ReportedSearchValues rootVals;
+    bool suc = getRootValues(rootVals);
+    if(!suc)
+      return false;
+    Player rootPla = getOpp(pla);
```
lightvector, line 1408: Shadows the member field `rootPla` and appears to be a bug imported from the old version of the code, flipping the player for no reason.

cpp/search/search.h
```diff
@@ -340,6 +342,13 @@ struct Search {
   //If node is not providied, defaults to using the root node.
   std::vector<double> getAverageTreeOwnership(int64_t minVisit, const SearchNode* node = NULL) const;
 
+  //Get ownership map as json
+  nlohmann::json Search::getJsonOwnershipMap(const Player pla, const Player perspective, const Board& board, const SearchNode* node, int ownershipMinVisits) const;
```
lightvector, line 346: There is an extra `Search::` here.

cpp/search/search.h
```diff
@@ -340,6 +342,13 @@ struct Search {
...
+  //Fill json with analysis engine format information about search results
+  bool Search::getAnalysisJson(const Player pla, const Player perspective, const Board& board, const BoardHistory& hist,
```
lightvector, line 348: Here too.

cpp/program/play.h
```diff
@@ -317,7 +317,8 @@ class GameRunner {
     Logger& logger,
     std::vector<std::atomic<bool>*>& stopConditions,
     std::function<NNEvaluator*()> checkForNewNNEval,
-    std::function<void(const Board&, const BoardHistory&, Player, Loc, const std::vector<double>&, const std::vector<double>&, const std::vector<double>&, const Search*)> onEachMove
+    std::function<void(const Board&, const BoardHistory&, Player, Loc, const std::vector<double>&, const std::vector<double>&, const std::vector<double>&, const Search*)> onEachMove,
+    bool logOwnership
```
lightvector, line 321: Should be named `alwaysIncludeOwnership` as well to be consistent with the implementation, and which is semantically more accurate.

cpp/search/searchresults.cpp
sanderland: Some header conflict which led to a 'i'll fix this later oversight'

cpp/program/play.h
sanderland: have named it includeOwnership in config and alwaysIncludeOwnership as var, to be consistent with the analysis cfg

sanderland:
Removed the pla parameter completely as it seemed redundant + fixes requested + other cleanup.

lightvector:
Merged, thanks!
Followed up with a little bit of minor refactoring and cleanup here: faf2525f
And some tests: 0d70dd58 1c877306

PR #381, opened 2020-12-31
Jp rules patch

lightvector:
Fixes a bug in the implementation of Japanese rules that would cause misevaluation of basic double ko death.

See discussion here of the issue:
https://lifein19x19.com/viewtopic.php?f=45&t=17963


lightvector:
Notably, **this adds a new kind of legal move**. It is now legal to move on top of an existing stone if that stone is in atari and marked as ko-recapture-blocked, as an alternative way of unblocking the ko recapture for a future turn. This is if player wants to perform a so-called "pass-for-ko" in Japanese-like postgame cleanup.

I'm a bit worried how the neural net will handle this, since currently moving on top of a stone is always illegal and probably the net has assigned it like -20 logits or something, so will be quite shocked when it sees the move is actually legal now. But hopefully this will still be okay and the net will adapt.

PR #384, opened 2021-01-10
Add nodeNumSymmetriesToSample option

roy7:
Earlier discussion found in https://github.com/roy7/KataGo/pull/1

The change to getHash() might break tests for nodes with symmetry. It shouldn't change behavior when symmetry is not used.


roy7:
May as well close it since it's so old. :)

PR #395, opened 2021-01-25
Better exceptions in distributed client

sanderland:
"Not 200" with the error in a long log is hard to parse and not so helpful. This changes the first line in the exception to have the returned status and error detail returned by the server.
In addition, the current behaviour makes it print to stderr on linux and stdout on windows I think, this PR changes windows to stderr.

This makes it so the first line has a fairly clear error message if available, and the errors can be picked up from stderr.
Exceptions are still a tad messy, but this allows reasonable feedback in a hypothetical frontend

image

image


lightvector:
Looks good, thanks! Testing this out soon, will also want to try it on windows with some exceptions to see if the output works for me as well.

PR #404, opened 2021-01-31
Fix buffer overrun in Tests::runBoardAreaTests()

zakki:
`NNInputs::fillScoring()` uses `Board::MAX_ARR_SIZE`, but tests use `NNPos::MAX_BOARD_AREA`


lightvector:
Great catch, thanks!

PR #425, opened 2021-02-22
converting to tf2

iopq:
I did some manual changes as well, but I also have a report
[report.txt](https://github.com/lightvector/KataGo/files/6022275/report.txt)


lightvector:
Thanks, this is great, will take a look this weekend or next.

lightvector:
Obviously, it's been a lot more than a week...
But I do hope to pick this up at some point. It might be some time more, but coming up near the top of the list after all the highthreads finishes is to try two new NN training ideas/projects, so that will be a good motivation to test this out and incorporate it. Thanks for making this PR available.

ChinChangYang:
It has been a long time until now.

Does this change pass any test cases in the most recent katago?

PR #426, opened 2021-02-22
Ensure performance reported

kcwu:
In some corner cases, ex. hit ctrl-c when all task loops are downloading models, performance would not be reported.


lightvector:
Looks good, thanks!

PR #430, opened 2021-02-26
Add psv to analyze commands

Dreamkeeper66666:


lightvector:
Thanks!

lightvector:
Sorry, hold on a moment - I may have been too quick to merge this. It seems it this PR doesn't update all the relevant places - for example it only adds the output to lz-analyze, and not kata-analyze.

I'm happy to just fix that up, but I wanted to also also ask, what's the need for this value? If the method of selecting which move to play changes, there is no guarantee right now that these values will remain similar - they could change around arbitrarily in relation to other things. Does your use case care about the exact values?

Dreamkeeper66666:
Sorry, I didn't learn the code well. Maybe I don't need this anymore. Btw, I just have a question. For the lcb part, are the selfutility and the radius^2 both weighting in visits? Thanks!

lightvector:
Selfutility is going to just be the average MCTS utility, viewed from the perspective of the current player. The radius is in units of utility as well, indicating the radius of the confidence interval of the utility around the selfutility.

Dreamkeeper66666:
> Selfutility is going to just be the average MCTS utility, viewed from the perspective of the current player. The radius is in units of utility as well, indicating the radius of the confidence interval of the utility around the selfutility.

Is it the weighted average or not?

lightvector:
Yes. It would be a pretty terrible utility estimate if each MCTS node used the unweighted average of all children beneath them. Then the search would hardly work at all. :)

Dreamkeeper66666:
> Yes. It would be a pretty terrible utility estimate if each MCTS node used the unweighted average of all children beneath them. Then the search would hardly work at all. :)

Got it, thanks! So are they weighting in the number of visits or something?

lightvector:
KataGo has a few modifications to how it computes utility that makes it not quite the same as just a weighted average by visits. I don't think you need to care about the exact details, unless you're doing something very particular and special? The utility estimate should just be considered a reasonable estimate of the value of the node that the search comes up with, given the limits of what the search is able to solve.

Dreamkeeper66666:
Thank you! I learned a lot.

PR #438, opened 2021-03-03
Friendlypass

lightvector:
Implements friendly passing in area scoring in GTP, if specifying rules that support it.


lightvector:
Ick, umm, let's retry that.

Issue #435, opened 2021-02-28
Can't open watchgame.txt in Windows due to "file is being used by another process" error.


cheeseonhead:
I can't open the watchgame.txt file while katago is running. Either using PowerShell's "Get-Content .\watchgame.txt -Tail 50 -Wait" function or trying to open the file in an text viewer.

l1t1:
ctrl c ctrl v to another file

cheeseonhead:
How do I do that when I can't open the file at all?

PR #441, opened 2021-03-03
Watch multiple games

cheeseonhead:
### Motivation
Sometimes an ongoing game gets boring so we should output multiple games so people can watch the most interesting one, or enable the watching app to pick.

### Proposal
Add an option called `watchOngoingGameMaxCount` that lets the user choose how many games they want in the watchGame file. In order to identify the games, additional output like game index and game id are also output.

Instead of the 0th thread writing to the file, now all threads whose index is less than `watchOngoingGameMaxCount` will output their moves.

Example new generated output:
```
Game #0 (A1CFF3FF47AFF5E4)
MoveNum: 6 HASH: F7F68A6B34005BEC6164319478DE21AD
   A B C D E F G H
12 . . . . . . . .
11 . . . . . . . .
10 . . . . O1. . .
 9 . . . O . X . .
 8 . . . . O3. . .
 7 . . . . X2@ . .
 6 . . . . . . . .
 5 . . . . . . . .
 4 . . . . . . . .
 3 . . . . X . . .
 2 . . . . . . . .
 1 . . . . . . . .

Network: rect15-b20c256-s343365760-d96847752
Rules: {"friendlyPassOk":false,"hasButton":false,"ko":"SIMPLE","komi":6.5,"scoring":"AREA","suicide":true,"tax":"NONE","whiteHandicapBonus":"0"}
Player: Black
Move: F7
Num Visits: 1200
Black Winrate: 47.0417%
Black Lead: 0.0853327
```

### Alternatives considered
Let each thread output to their own files, but then we need infrastructure to let the user set the output file pattern. Just seems like too much work.

### DIFF
--- a/cpp/command/contribute.cpp
+++ b/cpp/command/contribute.cpp
@@ -171,12 +171,14 @@ static void runAndUploadSingleGame(
   const string gameIdString = Global::uint64ToHexString(rand.nextUInt64());
 
   std::function<void(const Board&, const BoardHistory&, Player, Loc, const std::vector<double>&, const std::vector<double>&, const std::vector<double>&, const Search*)>
-    onEachMove = [&numMovesPlayed, &outputEachMove, &flushOutputEachMove, &logGamesAsJson, &alwaysIncludeOwnership, &gameIdString, &botSpecB, &botSpecW](
+    onEachMove = [&numMovesPlayed, &outputEachMove, &flushOutputEachMove, &logGamesAsJson, &alwaysIncludeOwnership, &gameIdx, &gameIdString, &botSpecB, &botSpecW](
       const Board& board, const BoardHistory& hist, Player pla, Loc moveLoc,
       const std::vector<double>& winLossHist, const std::vector<double>& leadHist, const std::vector<double>& scoreStdevHist, const Search* search) {
     numMovesPlayed.fetch_add(1,std::memory_order_relaxed);
     if(outputEachMove != nullptr) {
       ostringstream out;
+      out << "Thread: " << Global::intToString(gameIdx) << "\n";
+      out << "GameID: " << gameIDString << "\n";
       Board::printBoard(out, board, moveLoc, &(hist.moveHistory));
       if(botSpecB.botName == botSpecW.botName) {
         out << "Network: " << botSpecB.botName << "\n";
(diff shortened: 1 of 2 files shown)


lightvector:
Was https://github.com/lightvector/KataGo/issues/435 ever resolved?

I feel like given how crude all of this is, and given all the mess of file locking issues (and the horrible hack KataGo is using to work around Window's insistence on not flushing the file when you call flush) the way forward is probably more like if KataGo can have a more standard way to feed json data to an external GUI to display properly instead of ascii format in a text file, not unlike @sanderland implemented for KaTrain.

cheeseonhead:
No, I don't think #435 was ever resolved. I switched to linux just so I can build my little watcher app. The standard way sounds amazing but I would have no idea how to implement it.

Friday9i:
On my side, under Windows, I can't do anything with watchGame.txt: it is locked when KataGo is running, I can't even copy and paste the file to try to open it!
Note: I don't have any issues to open other KataGo files when KataGo is running (eg .sgfs and .log files when katago selfplay is running: I can open them, copy&paste, ... I just can't delete them, which is normal of course).

sanderland:
> The standard way sounds amazing but I would have no idea how to implement it.

There's the `logGamesAsJson` option, and then you basically just watch for lines from stdout starting with `{`

cheeseonhead:
> There's the `logGamesAsJson` option, and then you basically just watch for lines from stdout starting with `{`

Ok let me get that to display multiple games too

cheeseonhead:
@lightvector 
> I feel like … the way forward is probably more like if KataGo can have a more standard way to feed json data to an external GUI to display properly instead of ascii format in a text file, not unlike @sanderland implemented for KaTrain.

I reread your comment after @sanderland ‘s reply, it sounds like `logGameAsJson` is the infrastructure we want right? Or are you proposing an even more standard way to pass data?

Would I be able to change `logGameAsJson` to show multiple games and get this PR finished, or is this blocked on something bigger? I have no intentions of dealing with #435 in this PR by the way, I think that’s its own fix/hack

cheeseonhead:
Oh man I feel so dumb, the `logGameAsJson` is exactly what I want. Didn't even occur to me that that's an option when looking through the code.

I don't want to merge this PR anymore, closing.

PR #449, opened 2021-03-20
Seeming improvements to MCTS based on variance and uncertainty of things

lightvector:
Implements the three methods discussed in LZ&Friends discord chat here:
https://discord.com/channels/417022162348802048/583775968804732928/822568031552667709

More variations on these methods could be tested too, and I am unsure of the optimal values of the parameters.


lightvector:
The total gain, optimistically, might be like 80 Elo? But it seems smaller once multithreading gets involved. More testing needed!

====================================================================
Idea 1: Dynamic variance-scaled cPUCT 

* Keep a running estimate of variance of utility of playouts for each search node. Also mix in some sort of prior with some weight, to handle nodes with few playouts. 
* Scale cPUCT individually at every node of the search by sqrt(posterior variance estimate).

Motivation: 

The AlphaZero-style PUCT formula that drives MCTS is:
```
MCTS_Selection_Value(a) = Utility + cPUCT * Explore_Term(a)
where Explore_Term = Policy_Prior(N) sqrt(sum_b N_b) / (N_a + 1)
```
As in <https://colab.research.google.com/drive/14v45o1xbfrBz0sG3mHbqFtYz_IrQHLTg>, you can view the second term of the formula as a progressively widening confidence bound, or as in <https://arxiv.org/pdf/2007.12509.pdf> a regularization term from the prior policy. Also, we know that scaling up cPUCT very slightly as N increases is good.

But anyways, here, the key thing we notice is that `Explore_Term` is unitless, but it's being added to Utility, which is unitful, so cPUCT also needs to have units of utility, and needs to be scaled when the magnitude of utilities changes. What's mattering is the *scale* of the typical differences between utilities of nodes. If we double all of them by changing the units in which we express utility, cPUCT needs to double. For example, when converting from a [0,1] winrate to a [-1,1] advantage, the optimal cPUCT needs to double. 

However, the variance of utilities is also different in different game positions, or nodes of the search tree. So it stands to reason that generally scaling cPUCT by utility variance might be good. If utilities vary less than normal in some position, lowering cPUCT helps the search narrow in on the best branch. If utilities vary more than normal, raising cPUCT keeps the search from focusing on one move too fast - with high variance, it's more likely that alternatives could with more search turn out to be good.

Comments: 

Currently I coded up the prior as just a fixed value and weight, introducing two new parameters. Nothing fancy like trying to estimate it based on parent node, or past game states. Unfortunately I think now what I coded up is also a little mathematically buggy, needs a bit more work. Also, Kata isn't fully scaling by sqrt(empirical variance) - it's possible you might *not* want to, due the way that utility estimates have heavy tails. E.g. even if variance is near 0, the worry of a huge blind spot means that maybe you should still keep some minimum amount of explore.

At 1 thread, 500 playouts this was worth maybe something like 20 to 60 Elo.

This idea seems simple enough that there has to be *some* prior paper on it, right? But I'm having a hard time finding one.

====================================================================
Idea 2: Uncertainty-based weighting

* In training, record a TD-lambda-style value target. E.g. `tdvalue(t) = (1-lambda) ( v_t + lambda v_{t+1} + lambda^2 v_{t+2} + ... )`. KataGo is using nonzero lambda right now, but lambda = 0 is also possible. (Just record multiple lambdas. Each one is just one float to write down, super cheap).
* Train the neural net to predict *how wrong itself is* relative to this tdvalue target: 
```
predicted_value = output of value head after sigmoid or tanh activation
predicted_error = output of new error-predicting head after softplus activation

actual_error = (stop_gradient(predicted_value) - tdvalue)^2
# New loss function term
loss += coeff * loss_fn(actual_error, predicted_error)

loss_fn could itself be mse, or it could be something like huber loss
```
* So basically, the net is asked to predict the mean squared error of its own output relative to what MCTS of its slightly-future self will say.
* Everywhere in the search that uses visits, allow fractional visits and call it "weight" instead. The MCTS value of a node is the weight-averaged MCTS value of all children. PUCT formula uses weight everywhere it used to use visits.
* Instead of using a weight of 1.0 everywhere, use a weight of `min(cap, a * predicted_error ^ b)` for some parameters `cap, a, b`. Cap is so that if the net is extremely confident (e.g. an actually won/lost position) we don't get an absurd weight.

Motivation: 

KataGo has had this as an aux target for a while, but has never used it for anything. Recently I looked at it by hand, it's super cool and does everything you'd want. It's low in the early opening (despite the early opening having the maximum possible amount of variance left, because we're measuring variance to the near future, not variance to the end of the game). It's low in well-understood positions and decided endgames. It's somewhat high in the middle of fights. It's very high in the middle of ladders and forcing sequences and unresolved tactics.

So let's use it to search uncertain positions more deeply. Again from <https://colab.research.google.com/drive/14v45o1xbfrBz0sG3mHbqFtYz_IrQHLTg> we know that visits, i.e. sample size, acts like "confidence" or "precision" of the belief of a node. So we should decrease weight when the net itself thinks its error is high. And then, PUCT will search those nodes sooner and more. And making sure MCTS average values are also weighted by weight helps cancel out the degree to the extra exploration will pull down the MCTS values.

Comments: 

I'm sure there are more creative and possibly more effective ways to use it, but this seems like a simple approach.
At 1 thread, 500 playouts this was also worth maybe something like 20 to 60 Elo.

====================================================================
Idea 3: Weight capping by policy prior

When computing the MCTS value of a node, if a child would greatly bring down the value, AND that child's share of the weight at that node is MORE than the policy prior, cap its weight share to the policy prior (or some multiple of the policy prior).

Motivation:

When multithreading/batching, due to virtual loss or other thread exploration mechanisms, or due to variable weight per visit as in Idea 2 above, or due to simply FPU/PUCT, sometimes it may end up that, e.g. you have child A with 90% policy and 10 weight (visits) and child B with 0.1% policy and 1 weight (visit).

This is extremely common with multithreading or batched playouts, since the extra playouts will have to go *somewhere*, even if it turns out there isn't any remaining high-policy-mass child to go on, but less extreme cases also happen normally. With such a low policy, usually child B will be an extremely bad move and letting it have even 10% of the node weight will massively downward bias the MCTS value of the node.

From <https://arxiv.org/pdf/2007.12509.pdf> we know that the limiting distribution of MCTS playouts is like a regularized posterior policy, where the policy has been shifted from the prior to upweight good moves and downweight bad moves. If we got evidence that a move was bad (low value) certainly we shouldn't ever be weighting that move *more* heavily than the prior!

Comments: 

I've in the past noticed some rare positions where multithreaded KataGo behaves poorly when one important move initiates a long one-way street. Too many threads go down it, and the eval gets corrupted for a while. Setting the search to use fewer threads often makes the problem go away, as well as creeping up on that position more gradually from many turns away. 

Implementing idea 2 made the issue worse. Adding this change fixed the problem. Because I wanted to be conservative and minimally invasive on the search (I've had lots of failed experiments at making the search "sharper"), I actually cap the weight share of a bad child at double the policy, instead of just the policy.

The specific implementation I used tries to avoid anything more than linear cost. Whenever updating the MCTS value of a node:

* Iterate through each child in sorted prior policy order (KataGo always adds children in policy order, so we're always already sorted). 
* As you iterate, track the average weighted value of children so far, and the total policy mass so far and total weight so far. 
* As you accumulate each child, if its value is sufficiently lower than the average so far (we actually use a soft threshold here), (soft-)cap its weight to `2 * total_weight_so_far * child_policy / total_policy_mass_so_far`.

Measuring the Elo effect of this change has been hard, since testing search at many threads and with enough playouts is hard. So I don't have an Elo number. In all anecdotal cases (e.g. interactive Lizzie) found idea 2 making things worse, this change fixes it though.

lightvector:
UEC parameters are:
```
useUncertainty = true
uncertaintyExponent = 1.0
uncertaintyCoeff = 0.25
cpuctUtilityStdevPrior = 0.30
cpuctUtilityStdevPriorWeight = 2.0
cpuctUtilityStdevScale = 0.85
useNoisePruning = true
```

Again, I'm not sure what the best ones are, particularly at larger numbers of playouts, or of these methods continue to be good at larger numbers of playouts. But maybe this is a starting point for testers. :)

lightvector:
UEC params here. This should make a reasonable starting point for anyone trying to tune these parameters. The values used here are very likely improvable with better tuning, most of which was done at 1 thread 500 playouts.
```
useUncertainty = true
uncertaintyExponent = 1.0
uncertaintyCoeff = 0.25
cpuctUtilityStdevPrior = 0.30
cpuctUtilityStdevPriorWeight = 2.0
cpuctUtilityStdevScale = 0.85
useNoisePruning = true
```

sbbdms:
Some questions about testing parameters:
(1) If I am going for a test between 2 bots, then how many winrate & games they play, can make me conclude that one bot is better than the other?
(2) If I am going to test multiple (>2) parameter combinations of idea 2:
(2-1) Can I only set "useUncertainty = false" as a primary bot and "useUncertainty = true" with multiple parameter combinations as secondary bots (So that the secondary bots won't play against each other), or do I have to let multiple parameter combinations play against each other too?
(2-2) If I have to let multiple parameter combinations play against each other for hundreds of games, then how could I record the game results and calculate their elos conveniently?

Thanks!

lightvector:
1. You can do whatever you like, as long as you say precisely what your test conditions were and do the proper statistics and report the confidence interval on the result you get. :)
https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval
https://en.wikipedia.org/wiki/Elo_rating_system#Mathematical_details

2-1. Again, you can do whatever you like, as long as you say exactly what you did. Usually it's better to have more variety of opponents, of course, rather than everything comparing to only one thing.
If you are going to do any testing with multiple threads, you should probably NOT use katago match at all, however, because the very same caching that makes it efficient may also affect the threading behavior. In that case, you'll need to use gomill's ringmaster https://mjw.woodcraft.me.uk/gomill/doc/0.8/ringmaster.html or gogui twogtp or some other external match script, to run KataGo via gtp end-to-end.

2-2. https://www.yuzeh.com/etc/2019-04-07-bayeselo-for-games

lightvector:
Hmmm, that first wikipedia page is kind of terrible for anyone not already familiar with the math. But if you google around, you can find some calculators, like this one: https://sample-size.net/confidence-interval-proportion/   (e.g. N = total games, x = wins, CL is the desired confidence level).

And then to convert the upper and lower bounds to Elo difference:
Elo difference = log base 10 (probability of win / (1 - probability of win)) * 400

lightvector:
And another reference for BayesElo with sgfs: https://lifein19x19.com/viewtopic.php?p=237932#p237932

sbbdms:
Thanks for providing these links! It seems I have to write a custom program to extract result from *.sgfs files and output to a minimal PGN file. Though it might cost some time, it feels much better to record everything manually. I will have a try!

sbbdms:
Thanks for providing these links! It seems I have to write a custom program to extract result from *.sgfs files and output to a minimal PGN file. Though it might cost some time, it feels much better than recording everything manually. I will have a try!

sbbdms:
Hi.
I tested the idea 2 these days, below are my period and result. Could you please check whether it is convincible?

General settings:
(1) Using kata1-b40c256-s7217565440-d1751008619 network (The strongest network during the test)
(2) numSearchThreads = 1
(3) maxPlayouts = 500
(4) Japanese rule
(5) idea 3 is enabled, idea 1 is also enabled with default parameters
You can check other details in my config files below.

Firstly I select
(1) uncertaintyExponent = 0.95/1.00/1.05
(2) uncertaintyCoeff = 0.20/0.25/0.30
9 combinations as 9 secondary bots (Because I wanted to make the result easier to record manually, as I didn't know how to use BayesElo at that time), plus 1 original bot with setting useUncertainty = false, totally 10 bots.
I generated 640 games totally, with the setting above.

Later I select
(1) uncertaintyExponent = 0.90/0.95/1.00/1.05/1.10
(2) uncertaintyCoeff = 0.15/0.20/0.25/0.30/0.35
25 combinations as 25 bots (No secondary bots anymore), plus 1 original bot with setting useUncertainty = false, totally 26 bots.
I generated 2000 games totally, with the setting above. 6 games of them are going to be rejected by BayesElo, as their results are "Void".

I write a custom program to extract all the results from .sgfs files to .pgn files and use BayesElo to calculate.
According to the link you mentioned: https://lifein19x19.com/viewtopic.php?p=237932#p237932
I set offset 2000, advantage 0, drawelo 0.01.
If I calculate all the 2640 games, the result is:

> version 0056, Copyright (C) 1997-2007 Remi Coulom.
compiled Jan 30 2007 20:30:07.
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under the terms and conditions of the GNU General Public License.
See http://www.gnu.org/copyleft/gpl.html for details.
ResultSet>readpgn minimal.pgn
2634 game(s) loaded, 6 game(s) with unknown result ignored.
ResultSet>elo
ResultSet-EloRating>offset 2000
2000
ResultSet-EloRating>advantage 0
0
ResultSet-EloRating>drawelo 0.01
0.01
ResultSet-EloRating>mm
00:00:00,00
ResultSet-EloRating>exactdist
00:00:00,77
ResultSet-EloRating>ratings
Rank Name        Elo    +    - games score oppo. draws
   1 1.10+0.20  2048   54   53   146   58%  1999    5%
   2 1.05+0.25  2038   42   42   233   58%  1987    7%
   3 0.90+0.30  2037   53   52   153   56%  1999    3%
   4 1.00+0.15  2029   50   49   164   55%  1999    7%
   5 0.90+0.20  2026   52   51   147   54%  2000    9%
   6 0.90+0.35  2022   54   53   144   53%  1999    4%
   7 0.95+0.25  2012   43   42   218   54%  1986    9%
   8 1.05+0.30  2005   42   42   224   53%  1987    8%
   9 1.05+0.20  2005   42   42   226   52%  1988    7%
  10 0.90+0.15  2002   51   50   158   50%  1999    6%
  11 1.00+0.25  2001   43   42   225   52%  1986    5%
  12 1.10+0.35  2000   54   54   141   50%  2000    4%
  13 0.90+0.25  1998   49   49   164   50%  2001    7%
  14 1.05+0.35  1997   51   51   146   50%  2000   10%
  15 1.00+0.20  1996   44   44   215   52%  1986    5%
  16 0.95+0.15  1996   53   53   148   49%  2001    3%
  17 0.95+0.30  1996   43   43   216   51%  1988    6%
  18 1.10+0.30  1992   50   51   159   48%  2002    5%
  19 1.00+0.30  1990   41   41   232   50%  1987    8%
  20 0.95+0.35  1984   50   50   160   48%  1998    8%
  21 1.10+0.15  1982   50   50   162   47%  2002    6%
  22 1.00+0.35  1980   52   53   146   47%  1998    5%
  23 1.10+0.25  1976   50   50   158   46%  2001    8%
  24 0.95+0.20  1976   42   42   230   48%  1990    7%
  25 Original   1961   23   23   800   44%  2002    6%
  26 1.05+0.15  1952   51   52   153   42%  2001    7%
ResultSet-EloRating>

If I only calculate the 2000 games where all the 26 bots participate in with no secondary bots, the result is:

> version 0056, Copyright (C) 1997-2007 Remi Coulom.
compiled Jan 30 2007 20:30:07.
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under the terms and conditions of the GNU General Public License.
See http://www.gnu.org/copyleft/gpl.html for details.
ResultSet>readpgn minimal2.pgn
1994 game(s) loaded, 6 game(s) with unknown result ignored.
ResultSet>elo
ResultSet-EloRating>offset 2000
2000
ResultSet-EloRating>advantage 0
0
ResultSet-EloRating>drawelo 0.01
0.01
ResultSet-EloRating>mm
00:00:00,00
ResultSet-EloRating>exactdist
00:00:00,71
ResultSet-EloRating>ratings
Rank Name        Elo    +    - games score oppo. draws
   1 1.10+0.20  2049   54   52   146   58%  1999    5%
   2 0.90+0.30  2037   53   52   153   56%  1998    3%
   3 0.95+0.25  2029   52   51   148   55%  1997    9%
   4 1.00+0.15  2028   50   49   164   55%  1998    7%
   5 0.90+0.20  2025   52   51   147   54%  2000    9%
   6 1.05+0.25  2023   50   49   161   54%  1998    7%
   7 0.90+0.35  2022   54   53   144   53%  2000    4%
   8 1.00+0.30  2015   49   49   162   53%  1998    9%
   9 1.05+0.20  2014   51   51   154   52%  2001    8%
  10 1.00+0.25  2008   52   51   153   51%  1999    5%
  11 0.90+0.15  2002   51   50   158   50%  1999    6%
  12 1.05+0.30  2000   51   51   154   50%  2000    6%
  13 1.10+0.35  1999   54   54   141   50%  1999    4%
  14 0.95+0.30  1998   52   52   144   50%  2001    8%
  15 0.90+0.25  1998   49   49   164   50%  2000    7%
  16 1.05+0.35  1997   51   51   146   50%  1999   10%
  17 0.95+0.15  1996   53   53   148   49%  2001    3%
  18 1.10+0.30  1992   50   51   159   48%  2002    5%
  19 0.95+0.35  1985   50   50   160   48%  1999    8%
  20 1.10+0.15  1982   50   50   162   47%  2002    6%
  21 1.00+0.20  1982   53   53   145   48%  1998    6%
  22 1.00+0.35  1980   52   53   146   47%  1998    5%
  23 1.10+0.25  1977   50   50   158   46%  2002    8%
  24 0.95+0.20  1972   50   50   158   45%  2004    9%
  25 1.05+0.15  1952   51   52   153   42%  2001    7%
  26 Original   1938   50   51   160   40%  2001    8%
ResultSet-EloRating>

So it seems that uncertaintyExponent = 1.10, uncertaintyCoeff = 0.20 are the optimal parameters during all the 25 combinations, according to this test.

I have 3 questions to ask:
(1) Do you think the test is convincible? Any mistake I made in the setting?
(2) Since I don't really understand the parameters mathematically, are the "optimal parameters" above expected theoretically?
(3) Since I can't use discord conveniently, have other testers done similar tests? What are their conclusions?

If there are no mistakes during the test, I will continue to try a higher uncertaintyExponent, with uncertaintyCoeff fixed at 0.20, to see if other optimal parameters exist.

Below is all the files I made during the test, you can have a check if necessary.

Thanks!

[Test of idea 2 (20210323~20210325).zip](https://github.com/lightvector/KataGo/files/6203540/Test.of.idea.2.20210323.20210325.zip)

sbbdms:
Updating the latest test result. The optimal parameters has changed from 1.10+0.20 to 0.90+0.30. It seems 2640 games are still not enough from choicing the optimal parameters. Will go on updating if my test makes no mistake.

Latest update 2021-03-26

2640+640=3280 games (With the first 640 games where only 10 bots participate in, 9 secondary bots):

> version 0056, Copyright (C) 1997-2007 Remi Coulom.
compiled Jan 30 2007 20:30:07.
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under the terms and conditions of the GNU General Public License.
See http://www.gnu.org/copyleft/gpl.html for details.
ResultSet>readpgn minimal.pgn
3273 game(s) loaded, 7 game(s) with unknown result ignored.
ResultSet>elo
ResultSet-EloRating>offset 2000
2000
ResultSet-EloRating>advantage 0
0
ResultSet-EloRating>drawelo 0.01
0.01
ResultSet-EloRating>mm
00:00:00,00
ResultSet-EloRating>exactdist
00:00:00,74
ResultSet-EloRating>ratings
Rank Name        Elo    +    - games score oppo. draws
   1 0.90+0.30  2058   46   45   205   59%  1998    5%
   2 1.10+0.20  2049   46   45   202   58%  1999    6%
   3 1.05+0.25  2032   38   38   279   57%  1988    8%
   4 1.10+0.35  2020   46   46   190   53%  1998    7%
   5 0.90+0.20  2019   45   45   197   53%  2000    8%
   6 0.90+0.35  2018   47   46   188   53%  2000    5%
   7 1.00+0.15  2015   43   43   218   53%  1999    8%
   8 1.05+0.30  2013   39   39   262   54%  1987    8%
   9 0.95+0.25  2010   38   38   272   53%  1989    8%
  10 1.05+0.35  2002   45   45   194   51%  1999    9%
  11 1.00+0.25  1999   39   39   275   52%  1988    5%
  12 1.10+0.30  1999   45   45   203   50%  2000    6%
  13 0.90+0.15  1996   45   45   196   49%  2000    7%
  14 1.05+0.20  1994   39   39   268   51%  1990    6%
  15 1.00+0.20  1993   40   40   257   51%  1988    6%
  16 0.95+0.30  1992   39   39   263   50%  1989    6%
  17 1.00+0.35  1991   45   45   198   49%  1999    6%
  18 1.00+0.30  1989   38   38   284   50%  1989    7%
  19 0.95+0.35  1985   43   43   218   48%  2000    6%
  20 0.90+0.25  1984   44   44   206   47%  2002    6%
  21 1.10+0.25  1982   43   43   216   47%  2000    7%
  22 1.10+0.15  1982   44   44   212   47%  2002    6%
  23 0.95+0.20  1978   37   38   288   48%  1992    7%
  24 0.95+0.15  1978   45   45   204   47%  2001    3%
  25 1.05+0.15  1964   44   45   205   44%  2000    6%
  26 Original   1959   22   22   846   44%  2000    6%
ResultSet-EloRating>

2000+640=2640 games (Without the first 640 games):

> version 0056, Copyright (C) 1997-2007 Remi Coulom.
compiled Jan 30 2007 20:30:07.
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under the terms and conditions of the GNU General Public License.
See http://www.gnu.org/copyleft/gpl.html for details.
ResultSet>readpgn minimal2.pgn
2633 game(s) loaded, 7 game(s) with unknown result ignored.
ResultSet>elo
ResultSet-EloRating>offset 2000
2000
ResultSet-EloRating>advantage 0
0
ResultSet-EloRating>drawelo 0.01
0.01
ResultSet-EloRating>mm
00:00:00,00
ResultSet-EloRating>exactdist
00:00:00,71
ResultSet-EloRating>ratings
Rank Name        Elo    +    - games score oppo. draws
   1 0.90+0.30  2057   46   45   205   59%  1998    5%
   2 1.10+0.20  2049   45   45   202   58%  1999    6%
   3 0.95+0.25  2022   45   44   202   54%  1999    7%
   4 1.10+0.35  2019   46   46   190   53%  1997    7%
   5 0.90+0.20  2019   45   45   197   53%  2000    8%
   6 1.05+0.25  2019   44   43   207   53%  1998    9%
   7 0.90+0.35  2018   47   46   188   53%  2001    5%
   8 1.00+0.15  2015   43   43   218   53%  1999    8%
   9 1.05+0.30  2012   46   45   192   52%  1998    7%
  10 1.00+0.30  2008   43   43   214   52%  1998    8%
  11 1.00+0.25  2005   45   45   203   51%  1999    4%
  12 1.05+0.35  2001   45   45   194   51%  1998    9%
  13 1.10+0.30  2000   45   45   203   50%  2001    6%
  14 1.05+0.20  1999   45   45   196   49%  2001    7%
  15 0.90+0.15  1996   45   45   196   49%  2000    7%
  16 0.95+0.30  1994   46   46   191   49%  2000    8%
  17 1.00+0.35  1991   45   45   198   49%  1999    6%
  18 0.95+0.35  1985   43   43   218   48%  1999    6%
  19 0.90+0.25  1984   44   44   206   47%  2001    6%
  20 1.10+0.25  1983   43   43   216   47%  2001    7%
  21 1.10+0.15  1982   44   44   212   47%  2002    6%
  22 1.00+0.20  1981   46   46   187   47%  1999    7%
  23 0.95+0.15  1978   45   45   204   47%  2001    3%
  24 0.95+0.20  1976   43   43   216   46%  2003    7%
  25 1.05+0.15  1964   44   45   205   44%  2000    6%
  26 Original   1942   44   45   206   41%  2002    7%
ResultSet-EloRating>

lightvector:
Thanks for the data - yes it is good to have independent confirmation that these parameters are good and better than original, at 500 playouts, 1 thread.
As for tuning the parameters, yes, you need many games to get an accurate tuning. 

Here is a tip that you might not have realized about how to read the result:
```
Rank Name Elo + - games score oppo. draws
1 0.90+0.30 2058 46 45 205 59% 1998 5%
```
The "+ -" being equal to 46 and 45 means that the confidence bound on that Elo number is 2058-45 to 2058+46, which is 2013 to 2104.

Configuration 8 on your list is this one, with 2013 Elo.
```
8 1.05+0.30 2013 39 39 262 54% 1987 8%
```
Because the confidence bound goes from 2013 to 2104 Elo, roughly speaking, that means that for every single configuration currently between 2 and 8, we are not fully confident whether 1 is actually better than it or not. Does that make sense?

Within bayeselo, if you use the "los" command, it stands for "Likelihood of Superiority", you can also have it print out a matrix of the likelihood, assuming validity of the Elo model and all the data so far, that a given configuration is superior to each other configuration.

Of course, with a more detailed analysis, we might also gain more confidence beyond BayesElo's consideration of each configuration independently of each other, by taking into account that all of these come from a 2-dimensional parameter space whose strength probably varies smoothly over that space. But that would take a bit more math.

sbbdms:
Thanks for the hint. It is difficult to me to use further command in BayesElo, or some advanced mathematical analysis...
As you have pointed out, currently the 1st parameters are not fully confident to be better than the parameters between 2nd to 8th.
According to that, could I lessen the number of bots from 26 to 8 which place from 1st to 8th? That could generate quite more games per bot, but I am not sure whether it will make the test less accurate...

sbbdms:
I have generated ~6000 games in total (Details in the result updated above). From the last 480 games I start setting the 14th~26th bots as secondary bots, to make sure there's at least 1 strong bot (1st~13th) in a game, and to contract their confidence bounds quickly.
The 1st bot is still 0.90+0.30, but its elo gap and LoS to the 2nd bot (1.10+0.20) is getting smaller. I am not sure if the 1st bot will change from 0.90+0.30 to others, actually the elo gaps between all the bots are quite small, most of them are <=5 elo.
(1) According to the result, the confidence bound is ±32 when ~400 games are played, and ±20 when 1050 games are played. If the actual elo gap between most of the bots are <=5 elo, then idk how many games are needed, to determine the best bot... Are there any other definite condition to determine the best bot, or do I really have to generate dozens of times of games to make the confidence bound to ±5?
(2) I notice that your comment mentioned:

> I'm sure there are more creative and possibly more effective ways to use it, but this seems like a simple approach.

So I guess that there might be some more changes to the code in the future. If so, will my data still be valid? Will I have to give up all the previous test and do it again?
Looking forward to your advice to my test. Thanks!

lightvector:
Thanks again for running these tests!

Given you have so many versions testing such small differences in parameters, it's probably impossible to practically run enough games to distinguish the exact perfect parameters, you have to accept that there will be some uncertainty. That's almost always the case with statistics. But hitting the exact perfect parameters is also not what we care about, since we know that the parameters may also vary with future neural nets, may be different with different threads and hardware, etc. Mostly just getting to the right overall values is more important.

Looking broadly at your results, you may notice that a lot of the uncertaintyCoeff values that are small are near the bottom - almost all the 0.15s are in the bottom half. And many of the 0.30s are in the upper half. So this suggests actually that maybe testing larger values would be good - and actually recently in discord chat https://discord.gg/bqkZAz3 "petgo3" also posted some results suggesting good performance with uncertainty coeff as high as 0.6 or 0.7!

However, like you, he didn't find any great benefit from increasing or decreasing uncertaintyExponent a lot. So actually to reduce the number of combinations, it may be beneficial to simply fix uncertaintyExponent = 1 and test a wider range of uncertaintyCoeff. Actually, if it did turn out that uncertaintyExponent = 1 was ideal, then this would also simplify the code, since uncertaintyExponent = 1 is the "no-op" case in the code too.

Regardless of who is testing, having additional tests from other users is almost always useful, since having more information about different machines and playouts and threads still helps. What do you think about doing a coarser sweep and seeing if much higher uncertaintyCoeff is better?

lightvector:
As for the detailed testing, basically I guess my overall advice would be - we should focus less on trying to find the exact number 1 out of a whole ton of configurations, and more on trying to find the overall approximate right value, and then validate that value across a wider range of playouts and threads.

sbbdms:
Thanks for your information and analysis --- I didn't notice the inner hint from my data...
I make a very rough statistic --- Summing the elo/rank, group by different uncertaintyExponent and uncertaintyCoeff. The higher sum of elo is better, the lower sum of rank is better.

> uncertaintyExponent | 0.9 | 0.95 | 1 | 1.05 | 1.1
Sum of the elo (Higher is better) | 10049 | 9960 | 10023 | 9994 | 10011

uncertaintyCoeff | 0.15 | 0.2 | 0.25 | 0.3 | 0.35
Sum of the rank (Lower is better) | 9966 | 10004 | 10038 | 10027 | 10002

uncertaintyExponent | 0.9 | 0.95 | 1 | 1.05 | 1.1
Sum of the elo (Higher is better) | 47 | 94 | 47 | 72 | 65

uncertaintyCoeff | 0.15 | 0.2 | 0.25 | 0.3 | 0.35
Sum of the rank (Lower is better) | 93 | 65 | 51 | 49 | 67

It does seem to be a hint that higher uncertaintyCoeff tends to be better (Is uncertaintyCoeff=0.35 an exception?).
for uncertaintyExponent, it seems 0.90 and 1.00 are both good. I think I will select them both in the future test.
So I am going to start a new test, selecting uncertaintyExponent = 0.90/1.00, and uncertaintyCoeff >= 0.20.
Going to record new results below. Thanks for your help!

sbbdms:
So far I generated ~11500 games for idea 2. My result may be a bit different from petgo's: uncertaintyCoeff may be better around 0.4 than 0.6 or 0.7 for my general settings. I plan to go to test idea 1 after each of the top bots in my result play >500 games, and then do a comparison to the original bot & UEC bot with different playout.

Could you please point out how to set to disable idea 1, so that I can do a comparison? In idea 1 there's no variable like useUncertainty in idea 2 to directly disable it...

lightvector:
Set cpuctUtilityStdevScale = 0 to disable idea 1.
Although, keep in mind that if idea 1 is also good (as it seems to be), then idea 1 will be active as well in any practical configuration.

lightvector:
Thanks, it's very cool to see your results show that at least at 500 playouts under your settings, 1.00+0.25 is very close to optimal as well, and even within statistical uncertainty of optimal still. For release and for implementation and theoretical simplicity, we might go with exponent of 1.00 anyways.

Just this week, I started some tests myself of a much smaller number of parameter settings at much higher and more expensive settings of 5000 playouts per move. It will be noisy with many fewer games, but will give some more indicative evidence on whether higher uncertaintyCoeff values are also good at high playouts, when combined with everything else.

sbbdms:
Updating the latest test result (For idea 1 only).

General settings:
(1) Using kata1-b40c256-s7217565440-d1751008619 network (The same as the network of the test for idea 2)
(2) numSearchThreads = 1
(3) maxPlayouts = 500
(4) Japanese rule
(5) idea 3 is enabled, idea 2 is also enabled, with uncertaintyExponent = 0.90, uncertaintyCoeff = 0.40, which is placed at 1st after 13120 games of the test

Latest update 2021-04-16
2021-04-16 +880 games

lightvector:
I took your results and plotted them in a 2D grid:
image

The bolded and underlined ones are the ones that have 500-700 games, and the rest are ones that have around 200-300 games.

I think plotting them like this makes the level of noise much more visually apparent. 
Roughly, we can see there is seemingly a diagonal band of good values, slanted a little:
image

Between these values it is almost impossible to tell apart which ones are better or worse, relative to the noise level, probably anything in this band, within reason, will perform well. So I think 0.9+0.35, or 1.0+0.25 are both pretty reasonable choices. You may be starting to hit diminishing returns in how much more additional games will help reduce the noise - since the noise goes as sqrt(N), it will take quadratically more and more games for the same relative gain in accuracy.

It actually makes a lot of sense that there would be a diagonal band. If you use the implemented formula for setting node weight as a function of node uncertainty, here is are the curves you get for uncertaintyExponent 1 and varying uncertaintyCoeff:
image

And here are the curves for uncertaintyCoeff 0.3 and varying uncertaintyExponent:
image

You can see that increasing either one has a very similar effect in the middle of the weights, so it makes sense that if one is increased, the other should be decreased, and vice versa. The main difference is how they behave out in the tail, but presumably it's hard to collect enough data from just playing games to reliably tell what tail behavior is optimal.

Based on your data, I think I'd be happy to go with either 0.9+0.35 or the UEC value of 1.0+0.25 if exponent 1 is preferable which turns out surprisingly better than I had expected it to come out from your testing. What do you think?

lightvector:
By the way, Elo of 1966 for 0.95+0.25 is so bizarrely out of place, you may want to check for the config you used to test it, it may have a typo, because that's a pretty big outlier compared to the values nearby.

sbbdms:
Thanks for your detailed diagrams and explanation that makes the result clearer!
As you have pointed out, there seems to be a diagonal band in the grid, and it is expected theoretically (Which it is unclear to me previously, but much clearer with your curves T.T).
According to the result above, it may have better combinations when uncertaintyExponent becomes lower/higher and uncertaintyCoeff becomes higher/lower. However as you have pointed out --- The elo difference might be little, also it needs too many games to reduce the noise --- It have cost me much time (~ 3 weeks) to generate, and the period even become slower with the amount of bots increased.
So currently I would like to turn to test idea 1 instead. I am going to use similar period with the current 1st combination of idea 2 (0.90+0.40) to find the corresponding good combinations of idea 1.
Further plan: I will choice 2~3 combinations from each idea --- Totally 4~9 combinations, to make a tournament (With 500/2000/5000 playouts, and the latest strongest network). After selecting the best combination, I will make a comparison against the original bot and the UEC bot.
Actually there seems to be some better combinations in idea 1 than the default UEC combination (0.30+0.85). However the current noise is quite big, and I will post the result when the noise become smaller.
BTW: I will check later: If there's a typo or other mistakes happened to the 0.95+0.25 bot in idea 2. However I think its chance should be very small...

sbbdms:
So far I generated ~9000 games for idea 1, the result is recorded above. The gap between different parameters are even smaller than idea 2...
I will spend another ~3 hours to make some top parameters in idea 1 play ~400 games, then I am going to combine parameters from both 2 ideas, and make a tournament:

(1) Select 3 parameters from idea 2, with different uncertaintyExponent and uncertaintyCoeff (According the result after 13120 games, they are 0.90+0.40, 1.00+0.25, 1.10+0.20), and select 3 parameters from idea 1 --- Totally 9 bots, to play 2250 games (~500 games per bot), with the latest strongest network (b40s785 for now), and 500 playouts per move.

(2) Select top 4 bots from the 9 bots above, to play 400 games (~200 games per bot), with 2000 playouts per move.

(3) Select top 2 bots from the 4 bots above, to play 200 games (200 games per bot), with 5000 playouts per move.

I will choose the winner as the best bot, and make a comparison against the original bot and the UEC bot, with 500/2000/5000 playouts per move.

The plan above is expected to be done within ~2 weeks.

What do you think about the result above and my next plan? Are there any different results from discord?

lightvector:
@sbbdms - Your plan sounds pretty good. Thanks for the testing! A few thoughts:

* I would be a little cautious about making `cpuctUtilityStdevScale` more extreme than 0.85. It's possible that it's fine, but it's also possible that setting it too extreme could cause unexpected problems later that are not uncovered simply by direct testing, because the effect of setting this parameter to values like 0.95 or 1.0 is effectively to greatly increase the sharpness of the search when the neural net thinks the position is very certain. For example, overly-sharp search could theoretically could contribute to worse handicap game play, or result in rare blind spots in the deep endgame in positions that occur when analyzing human games but not self-play.
If you're going to test a value like 0.95 instead of sticking with 0.85 or more conservative values, and such a value wins the tests, we might also want to validate in some H2 or H3 games against Leela Zero, or some H6-H9 games against SAI, or against older KataGo nets, afterward.

* For idea 1, if you've tested a large number of combinations, then if you can do a plot like I did for idea 2, to look at the 2-dimensional pattern formed by the results, it might reveal some things. Forming such a pattern can help see what is likely noise and what is likely a real effect - for example how we noticed the diagonal pattern for idea 2. Also, forming such a plot can give you an idea if perhaps one of the parameters simply doesn't matter at all - if there is no consistent pattern at all (diagonal or otherwise) and it looks like just noise, then we can simply stop tuning that parameter and then we get an easier problem of having 1 fewer parameter to worry about.

sbbdms:
20210427 idea 1

Above is a plot of the result of idea 1 in a 2D grid.

Initially I selected cpuctUtilityStdevPrior as 0.20 - 0.40. During the test I have noticed that lower cpuctUtilityStdevPrior like 0.20 and 0.25 might cause a weak performance --- This is also shown by the plot above. So I added higher value --- 0.45 and 0.50 to the test, but they doesn't seem to be better than the middle value like 0.35 and 0.40...

For cpuctUtilityStdevScale, as you mentioned, higher value like 0.90 and 0.95 are theoretically dangerous. The plot shows that their performance are very unstable with different cpuctUtilityStdevPrior --- Just like a wave, however it does exist 2 pairs of combination with good elos.

0.85 seems to be both good and stable, with cpuctUtilityStdevPrior between 0.30 and 0.50.

To the tournament I mentioned above, I selected 0.35+0.90, 0.40+0.85, 0.30+0.95 from idea 1. Do you think I should keep 0.35+0.90 and 0.30+0.95 --- Or discard them and add some other pairs of combinations with cpuctUtilityStdevScale = 0.85? If the latter, then I would restart the ongoing tournament.

Thanks and looking forward to your reply!

lightvector:
Thanks for the results! Unless some unexpected consideration is discovered, I think I am happy to use your recommendation of 0.40+0.85 / 1.00+0.25 to be the default once 1.9.0 is released. :)

Regarding idea 3, I would be suspicious that your result that it makes things weaker in 1 thread is simply noise. If you test with more games, I suspect you will see the effect reduce or go away. Idea 3 should have almost no effect with only 1 thread, so if you observe a change, it is more likely to be your observed change that is noise rather than there actually being an effect.

sbbdms:
I finished an extra test:
(1) Selecting the new parameters, the UEC parameters and the original version, with idea 3 enabled/disabled, totally 6 bots
(2) Network: b40s809
(3) 1 thread, 500 playouts, 1500 games, ~500 games per bot

The result:

> version 0056, Copyright (C) 1997-2007 Remi Coulom.
compiled Jan 30 2007 20:30:07.
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under the terms and conditions of the GNU General Public License.
See http://www.gnu.org/copyleft/gpl.html for details.
ResultSet>readpgn minimal.pgn
1499 game(s) loaded, 1 game(s) with unknown result ignored.
ResultSet>elo
ResultSet-EloRating>offset 2000
2000
ResultSet-EloRating>advantage 0
0
ResultSet-EloRating>drawelo 0.01
0.01
ResultSet-EloRating>mm
00:00:00,00
ResultSet-EloRating>exactdist
00:00:00,03
ResultSet-EloRating>ratings
Rank Name                               Elo    +    - games score oppo. draws
   1 0.30+0.85 / 1.00+0.25 / Disabled  2044   25   25   500   58%  1991    6%
   2 0.40+0.85 / 1.00+0.25 / Enabled   2038   25   25   500   57%  1992    6%
   3 0.30+0.85 / 1.00+0.25 / Enabled   2018   25   25   500   53%  1996    6%
   4 0.40+0.85 / 1.00+0.25 / Disabled  2008   25   25   499   51%  1999    6%
   5 Original / Original / Disabled    1950   25   25   500   41%  2010    8%
   6 Original / Original / Enabled     1943   25   26   499   40%  2012    6%
ResultSet-EloRating>los
                                  0. 0. 0. 0. Or Or
0.30+0.85 / 1.00+0.25 / Disabled     61 90 96 99 99
0.40+0.85 / 1.00+0.25 / Enabled   38    84 93 99 99
0.30+0.85 / 1.00+0.25 / Enabled    9 15    70 99 99
0.40+0.85 / 1.00+0.25 / Disabled   3  6 29    99 99
Original / Original / Disabled     0  0  0  0    65
Original / Original / Enabled      0  0  0  0 34
ResultSet-EloRating>

Brief conclusion:
(1) For the new parameters (0.40+0.85), enabling idea 3 is 30 elos STRONGER than disabling idea 3
(2) For the UEC parameters (0.30+0.85), enabling idea 3 is 26 elos WEAKER than disabling idea 3
(3) For the original version, enabling idea 3 is 7 elos WEAKER than disabling idea 3
Confidence bound is around ±25.

I don't know how to assess this result... Could it still be regarded as noise?

lightvector:
Yes, it sounds like this is all still noise. One parameter shouldn't have such a crazy different effect as a result of slight numerical changes.

lightvector:
I am closing this PR now, since all of these changes are merged in, and the new parameters are ready to go shortly for next release. Thanks for all the hard testing work!

PR #467, opened 2021-04-26
Update nneval.cpp to avoid bomb pattern

petgo3:
In analogy to daggerhack, i found the bomb pattern (which apeared quite often in KGS in the last time).
Since there were serveral variations of the pattern, i use more field specifications (m=4 and 5). Documented in the code...
Bomb patterns appeared at least quite with 15b and latest 40b, so it might need some more training....

### DIFF
--- a/cpp/neuralnet/nneval.cpp
+++ b/cpp/neuralnet/nneval.cpp
@@ -585,7 +585,7 @@ static const int bombPattern[8][8] = {
   {0,0,0,1,4,4,4,4},
 };
 static bool bombMatch(const Board& board, Player nextPla, Loc& banned, int symmetry) {
-    for (int yi = 0; yi < 9; yi++) {
+    for (int yi = 0; yi < 8; yi++) {
         for (int xi = 0; xi < 8; xi++) {
             int y = yi;
             int x = xi;
--- a/cpp/neuralnet/nneval.cpp
+++ b/cpp/neuralnet/nneval.cpp
@@ -567,7 +567,50 @@ static bool daggerMatch(const Board& board, Player nextPla, Loc& banned, int sym
   }
   return true;
 }
+// 1 own stone
+// 2 opp stone
+// 0 empty
+// 3 banned
+// 4 any
+// 5 opp or empty
 
+static const int bombPattern[8][8] = {
+  {0,0,0,0,0,0,0,4},
+  {0,0,0,0,0,0,0,5},
+  {0,0,2,3,0,0,5,2},
+  {0,1,1,2,0,0,5,5},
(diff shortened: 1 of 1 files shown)


petgo3:
e.g. game of worstman: 2021/04/24 04:37
http://kgs.gosquares.net/index.rhtml.en?id=petgo2&id2=worstman&r=1&ty=2019&tm=10

lightvector:
What happens if you don't have this hack, but you massively increase scoreUtility? For example, dynamic and static score utility both up to 0.7, and winrate utility down to 0.5?

sbbdms:
> for (int yi = 0; yi < 9; yi++) {

Should it be

for (int yi = 0; yi < 8; yi++) {

?

petgo3:
@sbbdms thx!
another example with two "bombs": (in sgf format)
[bomb_rsun.txt](https://github.com/lightvector/KataGo/files/6382137/bomb_rsun.txt)

petgo3:
@lightvector: even with normal config, i can't reproduce with simply starting at the cutting move on 3-4. But fact is, i have seen this pattern in many games, often in more than one corner (mostly h3-h6 games)

lightvector:
The thing that's going on here is that KataGo is a little more uncertain than usual about that fight - and quite reasonably so - the tactical sequence with all the cuts is decently long, and the whole thing is within one liberty of escape. In a handicap game, every move loses anyways, so why not try a sequence that is slightly messy and hope the opponent messes up? The issue of course, is that humans have memorized the sequence. So a part of this seems to be an opponent-modeling issue.

It's hard for training to improve this, I think, because it needs to be enough training to actually memorize the shape and overcome the prior of "razor-thin fights that are 30 moves deep that come down to 1-liberty of difference are good things to try when you're 99% behind"... which is actually a very reasonable prior!

If you go for about 10k playouts, KataGo seems to recognize though in many of these examples that starting this fight, although 1% better winrate (still losing, but every other move loses too), is a bit worse in score, and worse the deeper you go. So that's why I'm curious - can you first see what happens if you set the score utlility very strong and winrate utility lower (and use more than 10k playouts) - does that make a difference?

It's possible we'll also end up needing to implement dynamic komi - it's been pretty cool how far we've gotten without it though. Lastly, I do have one other idea I want to try out, which I won't share just yet. But for starters - it would be useful to know how much it's fixed or not fixed with extreme score utility settings.

lightvector:
Hmmm, maybe hold off on the score utility thing until I try my other idea. Working on it. :)

petgo3:
Not necessarily a good idea (Discussed with lv). Closed.

PR #475, opened 2021-05-01
minor changes on trim a string

fuhaoda:
Very minor changes. Since we already checked

```
if (p2 == string::npos)
    return string();
```
So we do not need `if (p1== string::npos) p1 = 0`.
I also added white space string as \v\f into " \t\r\n\v\f" based on C++ reference. All are very minor changes.
This change has passed all the test in the test folder folder.

### DIFF
--- a/cpp/core/global.cpp
+++ b/cpp/core/global.cpp
@@ -288,13 +288,10 @@ string Global::chopSuffix(const string& s, const string& suffix)
 
 string Global::trim(const string& s)
 {
-  size_t p2 = s.find_last_not_of(" \t\r\n");
+  size_t p2 = s.find_last_not_of(" \t\r\n\v\f");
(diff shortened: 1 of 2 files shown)


fuhaoda:
Add some minor changes on the parse board function, so that we can allow more flexible space rule. I tried tests, it works fine. BTW: do we need to consider to add a move constructor and move assignment operator for board class

fuhaoda:
Please ignore my comments on the move constructor and move assignment constructor. I just tried it this evening and it didn't make a difference.  https://en.cppreference.com/w/cpp/language/copy_elision

lightvector:
I made the whitespace trim changes manually in a separate commit., those look good. Closing this PR since I don't think we care to make the board parsing changes.

PR #477, opened 2021-05-02
Variety and pattern bonuses

lightvector:
This PR implements an experimental mechanism to make KataGo prefer to play local patterns (9x9) that have not been played before in the same game in a different corner, or that have not been played before in recent games.

The simplest use case: specify a new parameter in the config like:
`avoidRepeatedPatternUtility = 0.005`
This will make KataGo penalize by 0.005 utility (similar to about 0.25% winrate) *every* move that it plays whose 9x9 surrounding pattern is the same as a move it played earlier in the game. This is cumulative, so for example, a long forced sequence with N matching moves will be penalized N times as heavily. Note that because the value is cumulative per-move, especially for handicap games where the utility differences are already very small, this parameter does not need to be large at all to have a significant effect. 

A more complex way to use is like this, where you can also specify an *external* directory of SGF files that KataGo should avoid matching, or to mimic.
For example, if you had an automated bot on OGS or KGS, plus a script that automatically downloaded games that the bot played and put them into a folder, then this should allow the bot to play differently in each game, avoiding doing the same thing as the previous game.
```
# Load sgf files from this directory when the engine is started (ONLY on startup, will not reload unless engine is restarted)
avoidSgfPatternDirs = path/to/directory/with/sgfs/
# Load the newest this many files
avoidSgfPatternMaxFiles = 100
# Optional - if specified pay attention only to moves that were made by players with this name.
# For example you can set it to the name that your bot's past games will show up as in the SGF, so that the bot will only avoid repeating
# moves that itself made in past games, not the moves that its opponents made.
avoidSgfPatternAllowedNames = my-ogs-bot-name1,my-ogs-bot-name2 
# Penalize this much utility per matching move.
# You can also set this negative if you instead want to make KataGo favor the SGF patterns instead of penalize it.
avoidSgfPatternUtility = 0.001
# Penalty is multiplied by this per each older SGF file, so that old sgf files matter less than newer ones.
avoidSgfPatternLambda = 0.90
# Ignore any moves in SGF files that occurred before this turn number.
avoidSgfPatternMinTurnNumber = 0
```
Additionally, if the SGF comments a move as `%SKIP%`, this specific move will be ignored as well (the comment should occur on the SGF node where the move has been placed on the board).

Perhaps there may be many more creative uses for this - for example to force KataGo to positively prefer a particular joseki to generate lots of self-play or match games with that joseki, or other kinds of uses. Beware though that setting values too large may lead to "weird" and bad behavior, because the value head was not trained to "know" about these penalties. For example, setting a high penalty on a move that is preferred may lead to Kata to waste ko threats and escalate situations to delay having to play that move, only to be forced to play it later anyways.

If this PR gets merged, the final API for this may vary from what is implemented in this PR, so this API should not be considered stable yet.


lightvector:
Along with highthreads2 this is merged into master.

lightvector:
API stayed the same, except that it was extended a little. If you want to specify multiple sets of external files in different directories with individually different parameters, you can do:

```
avoidSgf2PatternDirs = ...
avoidSgf2PatternMaxFiles = ...
avoidSgf2PatternAllowedNames = ...
avoidSgf2PatternUtility = ...
avoidSgf2PatternLambda = ...
avoidSgf2PatternMinTurnNumber = ...


avoidSgf3PatternDirs = ...
avoidSgf3PatternMaxFiles = ...
avoidSgf3PatternAllowedNames = ...
avoidSgf3PatternUtility = ...
avoidSgf3PatternLambda = ...
avoidSgf3PatternMinTurnNumber = ...
```

and so on.

Issue #473, opened 2021-04-30
Making a "Pass" in the beginning of a game, abnormal search result in the log


sbbdms:
Brief description:
I was testing a pair of new parameters of the highthreads2 branch (https://github.com/lightvector/KataGo/pull/449).
I opened a new game against HiraBot (6d) on KGS, where KataGo played as white, gave 6 stones and afforded 9.5 reverse komi (http://files.gokgs.com/games/2021/4/28/KataGo3090-HiraBot41.sgf).

Unexpectedly KataGo played a "Pass" for its 2nd move...
The corresponding log of the move:

> 2021-04-29 04:48:22+0800: MoveNum: 2 HASH: 5289C1362E2F7B0E6905B7EE3B6492A6
   A B C D E F G H J K L M N O P Q R S T
19 . . . . . . . . . . . . . . . . . . .
18 . . . . . . . . . . . . . . . . . . .
17 . . . . . . . . . . . . . . . . . . .
16 . . . X . . . . . . . . . . . X . . .
15 . . X . . . . . . . . . . . . . . . .
14 . . O . . . . . . . . . . . . . . . .
13 . . . . . . . . . . . . . . . . . . .
12 . . . . . . . . . . . . . . . . . . .
11 . . . . . . . . . . . . . . . . . . .
10 . . . X . . . . . . . . . . . X . . .
 9 . . . . . . . . . . . . . . . . . . .
 8 . . . . . . . . . . . . . . . . . . .
 7 . . . . . . . . . . . . . . . . . . .
 6 . . . . . . . . . . . . . . . . . . .
 5 . . . . . . . . . . . . . . . . . . .
 4 . . . X . . . . . . . . . . . X . . .
 3 . . . . . . . . . . . . . . . . . . .
 2 . . . . . . . . . . . . . . . . . . .
 1 . . . . . . . . . . . . . . . . . . .

> koPOSITIONALscoreAREAtaxNONEsui0whbNfpok1komi-9.5
> Time taken: 12.047
> Root visits: 26234
> New playouts: 25506
> NN rows: 20165
> NN batches: 381
> NN avg batch size: 52.9265
> PlayoutDoublingAdvantage: 2.75
> PV: pass pass
> Tree:
> : T -117.92c W -99.99c S -17.93c (-61.7 L -57.8) N   26234  --  pass pass
> ---White(^)---
[...]

PR #478, opened 2021-05-09
Fast terminal fix

lightvector:
I think this should fix https://github.com/lightvector/KataGo/issues/473, which is caused by terminal nodes of the search running too fast and getting a too-large number of playouts while other threads are blocked on results from the GPU.


lightvector:
Merged into master along with highthreads2.

PR #483, opened 2021-05-17
Mimalloc support

isty2e:
This pull request adds the support for [mimalloc](https://github.com/microsoft/mimalloc), a better memory allocator. It has by claimed by authors that this outperforms other allocators including TCMalloc, but whether this applies to KataGo as well is beyond me.

I have confirmed that this can be built and run on both Windows and *NIX systems. Due to the problem of mimalloc itself, `ld` cannot correctly search the shared library file, and `LD_LIBRARY_PATH` should be set properly (like `export LD_LIBRARY_PATH=/usr/local/lib/mimalloc-1.7/:$LD_LIBRARY_PATH`) before running KataGo. 

Presumably the official windows binary can be built with this, and `mimalloc-override.dll`, `mimalloc-redirect.dll`, and a corresponding license file can be provided.


lightvector:
Cool, thanks. Is there any way to make the LD library path thing part of the cmake setup?

isty2e:
> Cool, thanks. Is there any way to make the LD library path thing part of the cmake setup?

I am not sure, but I consider it a problem of mimalloc itself (https://github.com/microsoft/mimalloc/issues/399), and setting the environment variable is a part of installation. Or one can consider statically linking it instead of using shared libraries.

PR #486, opened 2021-05-23
Slightly changed calculateAreaForPla to improve efficiency

fuhaoda:
Some modifications in Benson's algorithm region building algorithm to improve efficiency.

Based on the testing (https://github.com/fuhaoda/GoProject/blob/improvedbenson/DrBombeExplore/benchmarksolutions.cpp)
It can improve efficiency and can reduce time from 10.95 seconds to 8.26 seconds in release mode.

Overall, the codes are very well written, and I have tried many small modifications without too much improvement but reduce the code readability, then I discarded those changes. The current changes are the most significant improvements without change the input/output and structures.  

Two changes.
1) add isVlenNonZero into regionBuild function to improve efficiency
2) use memset to initialize vector. The internal std::fill (the implementation is a iteration) or iteration based initialization is slower than memset when setting all the values are the same. 
3) Change the Benson's algorithm loop to reduce re-calculate vitalCountByPlaHead

### DIFF
--- a/cpp/game/board.cpp
+++ b/cpp/game/board.cpp
@@ -1984,27 +1984,23 @@ void Board::calculateAreaForPla(
   bool plaHasBeenKilled[MAX_PLAY_SIZE];
   memset(plaHasBeenKilled, false, sizeof(plaHasBeenKilled[0])*numPlaHeads);
 
+  //Zero out vital liberties by head
+  uint16_t vitalCountByPlaHead[MAX_ARR_SIZE];
+  for(int i = 0; i<numPlaHeads; ++i)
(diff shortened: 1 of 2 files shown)


lightvector:
Merged, but aside from some minor style changes, there is one major potential bug, and also one major fragile point that needed to be documented that was introduced by these changes. These are fixed and/or documented by two followup commits I added:

https://github.com/lightvector/KataGo/commit/760b235e4a4c1079226594c604b1c6172a56796c
https://github.com/lightvector/KataGo/commit/0cc093f0111a6639d817d28f69f8496a5b3c2d5f

We can discuss when we chat next.

Speedup looks good on both windows and linux in my testing. Thanks!

PR #488, opened 2021-05-27
Better root info in analysis

### LINKED ISSUE #487: rootInfo consistent with sub-tree in analysis engine
Currently, the rootInfo is heavily skewed in case of widerootnoise, and a bit skewed in normal cases, particularly on low visit counts.
As wider exploration by default is considered, what is the best way to add some kind of unbiased root info?
That is 'the statistics for the root node ignoring visits that would not have happened if the root were explored like a sub-tree'.

- Straight up replace rootInfo with this, breaking change
- Some kind of flag to enable this -> breaks older katago versions not expecting the flag or requires version checking.
- An extra return field -> Getting a bit messy, but maybe best?

Any pointers to existing code which already does this and can be re-used for the implementation is also appreciated.

sanderland:
Resolves  #487

This PR tries to make rootInfo consistent with subtree analysis.

This is done using:

* A parameter to getPlaySelectionValues called neverUseLcb which bypasses the lcb weighting code and just does cpuct based pruning.
* A new function in searchresults.cpp which calculates values based on this weighting, similar to recomputeNodeStats


sanderland:
Definitely not working, will debug a bit.

sanderland:
Parameter wasn't beingi picked up on, defaultParams being somehow not the default parameters.
rootnoise=1 9x9 1000 visits
before ced8e66 
{**"scoreLead":-1.4793242642047661**,"scoreSelfplay":-1.8746242198343235,"scoreStdev":10.649210448204407,"symHash":"1521075FDD85FFE1B074F776AFC4E49D","thisHash":"1521075FDD85FFE1B074F776AFC4E49D","utility":-0.36914534969312135,"visits":1007,"winrate":0.3288142485813965}

after
{**"scoreLead":-0.2283353305117771**,"scoreSelfplay":-0.34701212632094613,"scoreStdev":10.551285398765346,"symHash":"1521075FDD85FFE1B074F776AFC4E49D","thisHash":"1521075FDD85FFE1B074F776AFC4E49D","utility":-0.13757544601522512,"visits":1007,"winrate":0.4322882676401262}

no noise
{**"scoreLead":-0.2110051170223215,"**scoreSelfplay":-0.2868103194469904,"scoreStdev":10.375365186328366,"symHash":"1521075FDD85FFE1B074F776AFC4E49D","thisHash":"1521075FDD85FFE1B074F776AFC4E49D","utility":-0.1155622656212169,"visits":1007,"winrate":0.44267916659723194},"turnNumber":0}

looks fine now

lightvector:
Hmmm. I was thinking maybe this wouldn't touch `recomputeNodeStats` at all, and not mutate anything at all. Just purely be a top-level accessor function like many others in searchresults.cpp. Then you also don't need to add any toggleable search parameters. Is there a good reason to do it this way?

sanderland:
> Hmmm. I was thinking maybe this wouldn't touch `recomputeNodeStats` at all, and not mutate anything at all. Just purely be a top-level accessor function like many others in searchresults.cpp. Then you also don't need to add any toggleable search parameters. Is there a good reason to do it this way?

I was avoiding copy-pasting large blocks of code, but maybe this isn't too bad.

"rootInfo":{**"scoreLead":-0.18699477485373903**,"scoreSelfplay":-0.2529932720177927,"scoreStdev":10.377733320126849,"symHash":"1521075FDD85FFE1B074F776AFC4E49D","thisHash":"1521075FDD85FFE1B074F776AFC4E49D","utility":-0.10033975015111665,"visits":1007,"winrate":0.44942799660448285},"turnNumber":0}

lightvector:
Looks a lot better... except getPrunedRootValues is now no longer being used by anything? Need to put it back into being used by analysis?

sanderland:
> Looks a lot better... except getPrunedRootValues is now no longer being used by anything? Need to put it back into being used by analysis?

Hmm, strange that I didn't notice. it seems a make clean/make was needed to show it was missing.
8b6a0ae should fix it

lightvector:
Thanks. Doing a closer review: I think we want to also fill in a few remaining missing fields on the ReportedSearchValues.
See code in `bool Search::getNodeValues(const SearchNode& node, ReportedSearchValues& values) const`
```
  values.staticScoreValue = ScoreValue::expectedWhiteScoreValue(scoreMean,scoreStdev,0.0,2.0,rootBoard);
  values.dynamicScoreValue = ScoreValue::expectedWhiteScoreValue(scoreMean,scoreStdev,recentScoreCenter,searchParams.dynamicScoreCenterScale,rootBoard);
```
And this bit:
```
  //Clamp. Due to tiny floating point errors, these could be outside range.
  if(values.winLossValue < -1.0) values.winLossValue = -1.0;
  if(values.winLossValue > 1.0) values.winLossValue = 1.0;
  if(values.noResultValue < 0.0) values.noResultValue = 0.0;
  if(values.noResultValue > 1.0-abs(values.winLossValue)) values.noResultValue = 1.0-abs(values.winLossValue);

  values.winValue = 0.5 * (values.winLossValue + (1.0 - values.noResultValue));
  values.lossValue = 0.5 * (-values.winLossValue + (1.0 - values.noResultValue));

  //Handle float imprecision
  if(values.winValue < 0.0) values.winValue = 0.0;
  if(values.winValue > 1.0) values.winValue = 1.0;
  if(values.lossValue < 0.0) values.lossValue = 0.0;
  if(values.lossValue > 1.0) values.lossValue = 1.0;
```
Particularly for this second bit, copying this would be an unfortunate bit of code duplication, but I'm happy to factor it out myself if you like. Probably would proceed by defining some helper methods on ReportedSearchValues to compute these values based on the other ones in search.h, and then implementing those helpers using these chunks of code with implementations in searchresults.cpp

Anyways, this looks great! Feel free to make these fixes if you like, otherwise I will make them myself as I later test this and merge in the run-up to assembling the next release.

sanderland:
Fixes in 3f04d0ef as just a new constructor for ReportedSearchValues.
Any reason for not using std::clamp ?

lightvector:
I don't think so, using clamp would be fine.

sanderland:
> I don't think so, using clamp would be fine.

Apparently this requires c++17, which is not as standard as I thought, so will leave it as is.

lightvector:
This is *actually* also merged into master. After git squash and merge, I cherry-picked the squashed commit into master.

lightvector:
I regenerated and updated tests as well. Looks good, thanks! :)

PR #493, opened 2021-06-04
Add a python file to summarize game results in sgfs file format

fuhaoda:
Here is the python file. Please take a look, and let me know your thoughts. Then I can modify format.

### DIFF
--- a/python/summarize_sgfs.py
+++ b/python/summarize_sgfs.py
@@ -1,3 +1,4 @@
+import glob
 import itertools
 
 from sgfmill import sgf
@@ -38,11 +39,11 @@ class GameResultSummary:
 
     call it by other function:
     import summarize_sgfs
-    elo_prior = 3400
-    gameResultSummary = summarize_sgfs.GameResultSummary(elo_prior)
-    gameResultSummary.add_games(input_file_dir)
-    gameResultSummary.print_game_results()
-    gameResultSummary.print_elos()
+    elo_prior = 50
+    game_result_summary = summarize_sgfs.GameResultSummary(elo_prior)
+    game_result_summary.add_games(input_file_dir)
+    game_result_summary.print_game_results()
+    game_result_summary.print_elos()
   """
 
   def __init__(
@@ -53,38 +54,34 @@ class GameResultSummary:
 
     self._all_sgfs_files = set()
     self._all_sgf_files = set()
-    self._new_sgfs_files = set()
-    self._new_sgf_files = set()
-    self._warning = False
-    self._elo_prior = elo_prior
+    self._should_warn_handicap_komi = False
+    self._elo_prior = elo_prior #number of prior games to stablize the results.
 
   """ Public functions """
 
   def add_games(self, input_file_dir, search_subdir=False):
     """add files under input_file_dir into the results and duplicate files will not be added."""
-    self._add_game_file_names(input_file_dir, search_subdir)
-    self._add_new_games_to_result_dict()
+    new_files = self._add_game_file_names(input_file_dir, search_subdir)
+    self._add_new_games_to_result_dict(new_files)
 
   def add_a_game_file(self, input_file_name):
     """add a new game file (sgfs or sgf) into the results and duplicate file will not be added."""
-    self._add_a_game_file_name(input_file_name)
-    self._add_new_games_to_result_dict()
+    new_file = self._add_a_game_file_name(input_file_name)
+    self._add_new_games_to_result_dict(new_file)
 
   def clear(self):
     """Clear all the records and reset to empty sets and empty dictionaries"""
     self.results = {}  # a dictionary as {(black_player_name, white_player_name):Record}
     self._all_sgfs_files = set()
     self._all_sgf_files = set()
-    self._new_sgfs_files = set()
-    self._new_sgf_files = set()
-    self._warning = False
+    self._should_warn_handicap_komi = False
 
   def print_elos(self) -> elo.EloInfo:
     """Print estimated Bayes Elo scores centered as Elo prior"""
     elo_info = self._estimate_elo()
     print(elo_info)
     print(f"The prior is set as {self._elo_prior}, and the estimated Bayes Elo:")
-    if self._warning:
+    if self._should_warn_handicap_komi:
       warnings.warn("There are handicap games or games with komi < 5.5 or komi > 7.5")
     return elo_info
 
@@ -109,74 +106,89 @@ class GameResultSummary:
   """ Private functions starts here, and no more public functions below """
 
   def _add_game_file_names(self, input_file_dir, search_subdir):
+    # handle user specifying multiple sgf files to the script via glob on the terminal, for example foo/bar/*.sgf"
+    print(f"Searching and adding files in {input_file_dir}, please wait ...")
+    if input_file_dir.lower().endswith((".sgf", ".sgfs")):
+      files = glob.glob(input_file_dir)
+      for file in files:
+        self.add_a_game_file(file)
+      print(f"We found {len(files)} new sgfs/sgf files in after parse {input_file_dir}.")
+      return {"new_sgfs_files": [], "new_sgf_files": []}
+
     if not os.path.exists(input_file_dir):
       print(f"There is no directory under name {input_file_dir}")
-      return None
-    _files = []
+      return
+
+    files = []
     if (search_subdir):
       for (dirpath, dirnames, filenames) in os.walk(input_file_dir):
-        _files += [os.path.join(dirpath, file) for file in filenames]
+        files += [os.path.join(dirpath, file) for file in filenames]
     else:
-      _files = [os.path.join(input_file_dir, file) for file in os.listdir(input_file_dir)]
+      files = [os.path.join(input_file_dir, file) for file in os.listdir(input_file_dir)]
 
-    _new_sgfs_files = set([file for file in _files if file.split(".")[-1].lower() ==
+    new_sgfs_files = set([file for file in files if file.split(".")[-1].lower() ==
                            "sgfs"])
-    _new_sgf_files = set([file for file in _files if file.split(".")[-1].lower() == "sgf"])
+    new_sgf_files = set([file for file in files if file.split(".")[-1].lower() == "sgf"])
 
-    self._new_sgfs_files = _new_sgfs_files.difference(self._all_sgfs_files)
-    self._new_sgf_files = _new_sgf_files.difference(self._all_sgf_files)
+    #update the new_sgfs_files and sgf_files to remove duplication
+    new_sgfs_files = new_sgfs_files.difference(self._all_sgfs_files)
+    new_sgf_files = new_sgf_files.difference(self._all_sgf_files)
 
-    self._all_sgfs_files = self._all_sgfs_files.union(_new_sgfs_files)
-    self._all_sgf_files = self._all_sgf_files.union(_new_sgf_files)
-    print(f"We found {len(self._new_sgfs_files)} new sgfs files and {len(self._new_sgf_files)} new sgf files in the "
+    self._all_sgfs_files = self._all_sgfs_files.union(new_sgfs_files)
+    self._all_sgf_files = self._all_sgf_files.union(new_sgf_files)
+    print(f"We found {len(new_sgfs_files)} new sgfs files and {len(new_sgf_files)} new sgf files in the "
           f"search "
           f"directory {input_file_dir}.")
+    return {"new_sgfs_files": new_sgfs_files, "new_sgf_files": new_sgf_files}
 
   def _add_a_game_file_name(self, input_file_name):
+    new_sgf_files = []
+    new_sgfs_files =[]
     if not os.path.isfile(input_file_name):
       print(f"There is no file with name {input_file_name}.")
-      return None
+      return {"new_sgfs_files": new_sgfs_files, "new_sgf_files": new_sgf_files}
+
+    if input_file_name in self._all_sgfs_files or input_file_name in self._all_sgf_files:
+      print(f"File name {input_file_name} has been added before")
+      return {"new_sgfs_files": new_sgfs_files, "new_sgf_files": [new_sgf_files]}
+
     if input_file_name.split(".")[-1].lower() == "sgf":
-      self._new_sgf_files = {input_file_name}.difference(self._all_sgf_files)
-      if len(self._new_sgf_files) == 0:
-        print(f"File name {input_file_name} has been added before")
-        return None
-      else:
-        self._all_sgf_files = self._all_sgf_files.union({input_file_name})
+      new_sgf_files = [input_file_name]
+      self._all_sgf_files = self._all_sgf_files.union({input_file_name})
     elif input_file_name.split(".")[-1].lower() == "sgfs":
-      self._new_sgfs_files = {input_file_name}.difference(self._all_sgfs_files)
-      if len(self._new_sgfs_files) == 0:
-        print(f"File name {input_file_name} has been added before")
-        return None
-      else:
-        self._all_sgfs_files = self._all_sgfs_files.union({input_file_name})
+      new_sgfs_files = [input_file_name]
+      self._all_sgfs_files = self._all_sgfs_files.union({input_file_name})
     else:
       print(f"{input_file_name} is not sgf or sgfs file, no game was added")
-      return None
+      return
+    return {"new_sgfs_files": new_sgfs_files, "new_sgf_files": new_sgf_files}
 
-  def _add_new_games_to_result_dict(self):
+  def _add_new_games_to_result_dict(self, new_files):
     """add all sgfs files first"""
+    new_sgf_files = new_files["new_sgf_files"]
+    new_sgfs_files = new_files["new_sgfs_files"]
+
     idx = 1
-    for sgfs in self._new_sgfs_files:
+    for sgfs in new_sgfs_files:
       self._add_one_sgfs_file_to_result(sgfs)
       if (idx % 10 == 0):
-        print(f"Addedd {idx} files out of {len(self._new_sgfs_files)} sgfs files.")
+        print(f"Addedd {idx} files out of {len(new_sgfs_files)} sgfs files.")
       idx += 1
-    print(f"We have added additional {len(self._new_sgfs_files)} sgfs files into the results.")
+    #print(f"We have added additional {len(new_sgfs_files)} sgfs files into the results.")
 
     idx = 1
-    for sgf in self._new_sgf_files:
+    for sgf in new_sgf_files:
       self._add_one_sgf_file_to_result(sgf)
       if (idx % 10 == 0):
-        print(f"Added {idx} files out of {len(self._new_sgf_files)} sgf files.")
+        print(f"Added {idx} files out of {len(new_sgf_files)} sgf files.")
       idx += 1
-    print(f"We have added additional {len(self._new_sgf_files)} sgf files into the results.")
+    #print(f"We have added additional {len(new_sgf_files)} sgf files into the results.")
 
   def _add_one_sgfs_file_to_result(self, sgfs_file_name):
     """Add a single sgfs file. Each row of such sgf file contain a single game as a sgf string"""
     if not os.path.exists(sgfs_file_name):
       print(f"There is no SGFs file as {sgfs_file_name}")
-      return None
+      return
 
     with open(sgfs_file_name, "rb") as f:
       sgfs_strings = f.readlines()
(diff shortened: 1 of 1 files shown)


python/summarize_sgfs.py
```diff
@@ -0,0 +1,120 @@
...
+  """Summarize Go games in sgfs file format. Each row of sgfs game is a single game in a sgf string."""
+
+  def __init__(
+      self,
+      file_path
+  ):
+    self.file_path = file_path
+    self.sgf_files = []
+    self.players = {}
+    self.results = []
```
lightvector, line 16: Why not have this be a dictionary?

You could have a collections.defaultdict mapping tuples of (black_player_name, white_player_name) -> WinLossStats, 

where WinLossStats could be a dataclass https://docs.python.org/3/library/dataclasses.html that just stores the count of wins and losses, defaulting to 0 on construction.

Then you could just do something like `self.resullts[(pla_black,pla_white)].wins += 1`.

This means you can get rid of all of the indexing logic, all of the table resizing code, and the player info helper, and it also becomes easy to enable the calculation of the P1Advantage since we keep track of who played  black and who played white.

You also don't need to store the set of players. If you set it up like this, you can compute it on the fly like:
`players = set(itertools.chain(*((pla_black,pla_white) for (pla_black,pla_white) in self.results.keys())))`

python/summarize_sgfs.py
```diff
@@ -0,0 +1,120 @@
...
+    for i in range(n_pla-1):
+      for j in range(i+1,n_pla):
+        won = self.results[self.players[pla_names[i]]][self.players[pla_names[j]]]
+        lost = self.results[self.players[pla_names[j]]][self.players[pla_names[i]]]
+        total_games = won + lost
+        data.extend(elo.likelihood_of_games(pla_names[i], pla_names[j], total_games, won/total_games, False))
+
+    data.extend(elo.make_center_elos_prior(pla_names, 0))
+    info = elo.compute_elos(data, verbose=True)
+    print(info)
```
lightvector, line 57: By default, we should probably output a warning if at least one of the sgfs are a handicap game. And output a warning if the komi is outside of the range from 5.5 to 7.5 *and* more than one komi value is used collectively among the sgfs. Basically, if the games played are unfair, and different games use very different settings, this Elo script is not going to produce meaningful results, since the assumption is that the games are fair, and users who accidentally try to use it should be warned.

python/summarize_sgfs.py
```diff
@@ -0,0 +1,120 @@
...
+      self.players[pla_name]= players_len
+      self.expand_result_table()
+
+
+if __name__ == "__main__":
+  description = """
+  Summarize SGFs files and estimate Bayes Elo score for each of the player.
+  """
+  parser = argparse.ArgumentParser(description=description)
+  parser.add_argument('-sgfs-input-dir', help='sgfs files input directory', required=True)
```
lightvector, line 114: Maybe this should be a positional argument with nargs = "+"
https://docs.python.org/3/library/argparse.html#nargs
so that users can specify multiple different directories.

Also as mentioned above, would be nice to have an argument to set the strength of a prior.

lightvector (commented):
Here are some comments! Ping me if you'd like to chat or have questions!

python/summarize_sgfs.py
```diff
@@ -0,0 +1,120 @@
...
+
+  def get_sgfs_files(self):
+    """Load all the sgfs file names in file_path directory into a list"""
+    if not os.path.exists(self.file_path):
+      print(f"There is no directory under name {self.file_path}")
+      return None
+    self.sgf_files = os.listdir(self.file_path)
+
+    for file in self.sgf_files:
+      if file.split(".")[-1].lower() != "sgfs":
```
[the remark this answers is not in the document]
fuhaoda: Very good suggestion. I just made the changes with more user friendly API. In addition, we can add both sgfs file from KataGo, and also sgf file as regular games. In addition, we didn't treat sgf file as a single line for a game. We read the whole sgf file as a game. But for sgfs file from katago, we assume that one line one game (sgf).

python/summarize_sgfs.py
```diff
@@ -0,0 +1,120 @@
...
+    self.results = []
+    self.get_sgfs_files()
+    self.parse_all_sgfs_files()
+
+  def get_sgfs_files(self):
+    """Load all the sgfs file names in file_path directory into a list"""
+    if not os.path.exists(self.file_path):
+      print(f"There is no directory under name {self.file_path}")
+      return None
+    self.sgf_files = os.listdir(self.file_path)
```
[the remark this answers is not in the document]
fuhaoda: Based on this, we just updated feature to add a directory, and also add a single sgf or sgfs file. In addition, we also allow user to add files in multiple steps, and clear() function to reset.

python/summarize_sgfs.py
```diff
@@ -0,0 +1,120 @@
...
+        self.sgf_files.remove(file)
+    print(f"Total {len(self.sgf_files)} sgfs files have been found under directory {self.file_path}.")
+
+
+  def parse_all_sgfs_files(self):
+    """Parse all sgfs files"""
+    idx=1
+    for file in self.sgf_files:
+        self.parse_one_sgfs_file(file)
+        print(f"Parsed {idx} files out of {len(self.sgf_files)} files.")
```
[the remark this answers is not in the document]
fuhaoda: Now we set as every 10 files instead of every line. Every 10 files will take about 10 seconds to parse large sgfs file.

python/summarize_sgfs.py
```diff
@@ -0,0 +1,120 @@
...
+    pla_names = list(self.players.keys())
+    n_pla = len(pla_names)
+    for i in range(n_pla-1):
+      for j in range(i+1,n_pla):
+        won = self.results[self.players[pla_names[i]]][self.players[pla_names[j]]]
+        lost = self.results[self.players[pla_names[j]]][self.players[pla_names[i]]]
+        total_games = won + lost
+        data.extend(elo.likelihood_of_games(pla_names[i], pla_names[j], total_games, won/total_games, False))
+
+    data.extend(elo.make_center_elos_prior(pla_names, 0))
```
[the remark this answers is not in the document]
fuhaoda: Based on this suggestion, we have made the prior input in constructor, and it can also input from command line.  We also add the print Elo prior before printing the result.

python/summarize_sgfs.py
fuhaoda: Very good suggestion. Now it has changed to a dictionary, and also use `players = set(itertools.chain(*(pla_names for pla_names in self.results.keys())))` to get player information. Remove the indexing log and resizing codes.

python/summarize_sgfs.py
fuhaoda: Added the check for handicap and Komi, and provide a warning if there are unfair game.

python/summarize_sgfs.py
fuhaoda: Done!

python/summarize_sgfs.py
```diff
@@ -0,0 +1,295 @@
...
+
+    self._add_a_single_sgf_string(sgf)
+
+  def _add_a_single_sgf_string(self, sgf_string):
+    """add a single game in a sgf string save the results in self.results."""
+    game = sgf.Sgf_game.from_bytes(sgf_string)
+    winner = game.get_winner()
+    pla_black = game.get_player_name('b')
+    pla_white = game.get_player_name('w')
+    if (game.get_handicap() is not None) or game.get_komi() < 5.5 or game.get_komi() > 7.5:
```
lightvector, line 204: Probably we should tolerate game.get_handicap() == 0?

python/summarize_sgfs.py
```diff
@@ -0,0 +1,295 @@
...
+    """
+    return self.results
+
+  """ Private functions starts here, and no more public functions below """
+
+  def _add_game_file_names(self, input_file_dir, search_subdir):
+    if not os.path.exists(input_file_dir):
+      print(f"There is no directory under name {input_file_dir}")
+      return None
+    _files = []
```
lightvector, line 115: Here and several variables below - no need to prefix with underscore for local variables, unlike object fields these should automatically be local/private.

python/summarize_sgfs.py
```diff
@@ -0,0 +1,295 @@
...
+
+  def __init__(
+      self,
+      elo_prior: float
+  ):
+    self.results = {}  # a dictionary as {(black_player_name, white_player_name):Record}
+
+    self._all_sgfs_files = set()
+    self._all_sgf_files = set()
+    self._new_sgfs_files = set()
```
lightvector, line 56: I think you can delete the `_new_sgfs_files` and `_new_sgf_files` fields too.
Usually, good programming practice involves minimizing the amount of nonlocal state that is maintained and minimizing the amount of code that depends on side effects.

We can change `_add_new_games_to_result_dict` to take an argument, the list of new games to add.

Then all of the game adding functions, rather than setting `_new_sgfs_files` and `_new_sgf_files` as a side effect, and then relying on someone to call `_add_new_games_to_result_dict`, should instead simply call `_add_new_games_to_result_dict` and pass it the new games to add. Now, a lot more of the data is passed by argument, and not by side effects.

python/summarize_sgfs.py
```diff
@@ -0,0 +1,295 @@
...
+    self._new_sgf_files = set()
+    self._warning = False
```
lightvector, line 58: Very minor: `_warning` could be better named `_should_warn_handicap_komi` or something like that.

python/summarize_sgfs.py
```diff
@@ -0,0 +1,295 @@
...
+
+
+class GameResultSummary:
+  """
+  Summrize Go games results in sgf or sgfs file format under a directory (and its subdirectory).
+  The sgfs file is a special file format provided by Katago, and each row is a single game in a sgf string.
+  sgf file is not required to be in a single line.
+
+  Public functions include:
+  self.add_games(input_file_dir, search_subdir = False): add files into the results and duplicate files will not be
```
lightvector, line 27: I had not thought of duplication checking. Very nice.

python/summarize_sgfs.py
```diff
@@ -0,0 +1,295 @@
...
+    print(f"We found {len(self._new_sgfs_files)} new sgfs files and {len(self._new_sgf_files)} new sgf files in the "
+          f"search "
+          f"directory {input_file_dir}.")
+
+  def _add_a_game_file_name(self, input_file_name):
+    if not os.path.isfile(input_file_name):
+      print(f"There is no file with name {input_file_name}.")
+      return None
+    if input_file_name.split(".")[-1].lower() == "sgf":
+      self._new_sgf_files = {input_file_name}.difference(self._all_sgf_files)
```
lightvector, line 140: Once you no longer have _new_sgf_files field, you can replace this line and the following line with:

```if input_file_name in self._all_sgf_files:```

and similar for other places where you test membership of just one file.

File: python/summarize_sgfs.py [hunk 14]
```diff
@@ -0,0 +1,295 @@
...
+    elif (winner == 'w'):
+      self.results[(pla_black, pla_white)].lost += 1
+    else:
+      self.results[(pla_black, pla_white)].draw += 1
+
+  def _estimate_elo(self) -> elo.EloInfo:
+    """Estimate and print elo values. This function must be called after add all the sgfs/sgf files"""
+    pla_names = set(itertools.chain(*(name_pair for name_pair in self.results.keys())))
+    data = []
+    for pla_Balck in pla_names:
```
lightvector, line 221: Black is misspelled Balck.

Also generally stick to lowercase for variable names, so `pla_black` and `pla_white`.

python/summarize_sgfs.py
```diff
@@ -0,0 +1,295 @@
...
+                      default=False)
+  parser.add_argument('-elo-prior', help='A Bayesian prior that the mean of all player Elos is the specified Elo ',
+                      required=False, type=float, default=0)
+  args = vars(parser.parse_args())
+
+  input_dirs = args["input_dirs"]
+  search_subdir = args["search_subdir"]
+  elo_prior = args["elo_prior"]
+
+  gameResultSummary = GameResultSummary(elo_prior)
```
lightvector, line 290: game_result_summary

python/summarize_sgfs.py
```diff
@@ -0,0 +1,295 @@
...
+      for pla_White in pla_names:
+        if (pla_Balck == pla_White):
+          continue
+        else:
+          record = self.results[(pla_Balck, pla_White)]
+          total = record.win + record.lost + record.draw
+          win = record.win + 0.5 * record.draw
+          winrate = win / total
+          data.extend(elo.likelihood_of_games(pla_Balck, pla_White, total, winrate, False))
+    data.extend(elo.make_center_elos_prior(list(pla_names), self._elo_prior))
```
lightvector, line 231: Sorry, I probably wasn't clear. :)

I didn't mean that you should be able to set the center of the elo prior. I meant that you should have an argument that controls the weight of a prior that for each player, pulls that player's Elo towards 0.

So, loop over all players and extend data with `elo.make_single_player_prior` with `num_games` based on the user's input. This will provide actual regularization of the Elos for players with too little data, not just center them.

python/summarize_sgfs.py
```diff
@@ -0,0 +1,295 @@
...
+
+if __name__ == "__main__":
+  description = """
+  Summarize SGF/SGFs files and estimate Bayes Elo score for each of the player.
+  """
+  parser = argparse.ArgumentParser(description=description)
+  parser.add_argument('-input-dirs', help='sgf/sgfs files input directories. If multiple directories, seperate them '
+                                          'with a space', required=True,
+                      nargs='+')
+  parser.add_argument('-search-subdir', help='If we also need to search all sub-directories', required=False,
```
lightvector, line 280: Just a question - is there a specific reason for making this an option, instead of just always searching all the subdirectories?

python/summarize_sgfs.py
```diff
@@ -0,0 +1,295 @@
...
+      results_matrix.append(row)
+    return results_matrix
+
+
+if __name__ == "__main__":
+  description = """
+  Summarize SGF/SGFs files and estimate Bayes Elo score for each of the player.
+  """
+  parser = argparse.ArgumentParser(description=description)
+  parser.add_argument('-input-dirs', help='sgf/sgfs files input directories. If multiple directories, seperate them '
```
lightvector, line 277: Make this a positional argument, one that doesn't require a flag?

python/summarize_sgfs.py
```diff
@@ -0,0 +1,295 @@
...
+      results[(black_player_name, white_player_name)].draw
+    """
+    return self.results
+
+  """ Private functions starts here, and no more public functions below """
+
+  def _add_game_file_names(self, input_file_dir, search_subdir):
+    if not os.path.exists(input_file_dir):
+      print(f"There is no directory under name {input_file_dir}")
+      return None
```
lightvector, line 114: Here and everywhere below: in a function that doesn't return anything, the normal convention is to just `return` rather than `return None`.

`return None` is normally used in functions that are capable of sometimes returning something besides `None`.

python/summarize_sgfs.py
```diff
@@ -0,0 +1,295 @@
...
+    _files = []
+    if (search_subdir):
```
lightvector, line 116: Additional condition. If the input_file_dir is not a directory but is in fact itself an sgf or sgfs file:

```
if os.path.isfile(input_file_dir) and input_file.lower().endswith(".sgf"):
  ...
elif os.path.isfile(input_file_dir) and input_file.lower().endswith(".sgfs"):
  ...
```
That way it should automatically handle user specifying multiple sgf files to the script via glob on the terminal, for example "foo/bar/*.sgf".

lightvector (commented):
Looks good. A few more comments that hopefully you also find useful or educational.

python/summarize_sgfs.py
fuhaoda: We parse the sgf string based on sgfmill package. In the source code when handicap is equal to 0, it will return to None. So our case has covered equal to 0 case. 

   ```
 def get_handicap(self):
        ...
        if handicap == 0:
            handicap = None
        elif handicap == 1:
            raise ValueError
        return handicap
```

File: python/summarize_sgfs.py (hunk 17 above)
Comment (fuhaoda) [reply python/summarize_sgfs.py:280]: In the 3rd page of the paper on MM algorithm for Bayes Elo calculation. It required that the players have to play with each other with win and lost, otherwise, the algorithm can't calculate the Elo score. Sometimes, if they keep searching all the subdirectories, it may accidentally include some games with players who didn't play with all the others, and only played with with a group, and this group is isolated. So the calculation will report errors. So we put `search-subdir` as false unless users know what they want to do.

File: python/summarize_sgfs.py (hunk 18 above)
Comment (fuhaoda) [reply python/summarize_sgfs.py:277]: save reason as above.

File: python/summarize_sgfs.py (hunk 9 above)
Comment (fuhaoda) [reply python/summarize_sgfs.py:115]: Agreed. Thanks for this comment

File: python/summarize_sgfs.py (hunk 11 above)
Comment (fuhaoda) [reply python/summarize_sgfs.py:58]: Agreed. Changed.

File: python/summarize_sgfs.py (hunk 12 above)
Comment (fuhaoda) [reply python/summarize_sgfs.py:27]: Thanks :)

File: python/summarize_sgfs.py (hunk 14 above)
Comment (fuhaoda) [reply python/summarize_sgfs.py:221]: checked and changed all

File: python/summarize_sgfs.py (hunk 15 above)
Comment (fuhaoda) [reply python/summarize_sgfs.py:290]: Agreed and changed.

File: python/summarize_sgfs.py (hunk 19 above)
Comment (fuhaoda) [reply python/summarize_sgfs.py:114]: Changed them all and agreed. This change will make the style consistent. But it seems that they are the same in machine code level https://stackoverflow.com/a/15300733/6598108

File: python/summarize_sgfs.py (hunk 10 above)
Comment (fuhaoda) [reply python/summarize_sgfs.py:56]: Good suggestion. I agreed. The `_new_sgfs_files` should not be meaningful in the global level and it increased risk of misuse. I just changed them as argument.

File: python/summarize_sgfs.py (hunk 13 above)
Comment (fuhaoda) [reply python/summarize_sgfs.py:140]: Agreed and changed.

File: python/summarize_sgfs.py (hunk 17 above)
Comment (lightvector) [reply python/summarize_sgfs.py:280]: This sort of makes sense, but why do you have directories where including all the sgfs in the directory are ones you want to include, but somehow you have further subdirectories that include extraneous data that should not be included? And you specifically rely on the extraneous files to be in a subdirectory rather than in the top-level directory to be included? This seems like a confusing way to organize your SGF files.

File: python/summarize_sgfs.py (hunk 16 above)
Comment (fuhaoda) [reply python/summarize_sgfs.py:231]: Below is the changes based my understanding on this comment. Please see if this change is right. This might be the one that I am not 100% sure (but 80% sure) on comment. The propose of the prior is to stabilize small number of games having big impact, and we still need to center the Elo at 0. So the changes will be,

   ```
 for pla in pla_names:      
      data.extend(elo.make_single_player_prior(pla, self._elo_prior,0))      
    data.extend(elo.make_center_elos_prior(list(pla_names), 0))
```

python/summarize_sgfs.py
lightvector: This one I don't understand the reasoning. How does making this require a flag help against including SGF files? The only difference is whether the user has to type `-input-dirs` or not before typing the list of files.

lightvector: against including *mistaken* SGF files.

python/summarize_sgfs.py
lightvector: This seems fine to me. Adding a 0 prior to all players will make it so that make_center_elos_prior is redundant, because if all players individually have a 0 prior, the collective error will also be minimized when their average is 0. But it also does no particular harm to have make_center_elos_prior as well.

python/summarize_sgfs.py
fuhaoda: Very good suggestion. This function has been added.

python/summarize_sgfs.py
fuhaoda: I agreed that is redundant. Let's remove `data.extend(elo.make_center_elos_prior(list(pla_names), 0))`.

python/summarize_sgfs.py
fuhaoda: That is correct comment. When I was trying to debug my codes and test the subdirectories function, I copied some other sgfs files into a subdirectory and found this error. So we made the search-sub default as false.

python/summarize_sgfs.py
fuhaoda: I was thinking to be more specific. Let me know if you prefer other way, we can modify it. I am absolutely flexible here.

PR #496, opened 2021-06-07
Make smaller board max lengths compatible

iopq:
Can decrease RAM usage

### DIFF
--- a/cpp/command/gtp.cpp
+++ b/cpp/command/gtp.cpp
@@ -409,8 +409,8 @@ struct GTPEngine {
 
     bool wasDefault = false;
     if(boardXSize == -1 || boardYSize == -1) {
-      boardXSize = 19;
-      boardYSize = 19;
+      boardXSize = Board::MAX_LEN;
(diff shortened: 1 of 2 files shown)


lightvector:
These are the only places needed to make it work? Cool, thanks. Can we define a new constant `Board::DEFAULT_LEN` for these spots? On the side where we compile for support for larger sizes, rather than smaller, probably KataGo shouldn't quite *default* to be a larger size than 19 even if it supports it.

iopq:
This only makes the part that I tested work, most of the tests do not work, of course

I can change it to separate default length and max length, of course

lightvector:
I dehardcoded 19s on the spots you highlighted and a few other spots I thought were reasonable: https://github.com/lightvector/KataGo/commit/516181f96c170eacbfd7bad8e1d4d6e7a4b2a486

And I did it via a new variable that should automatically shrink if the max len is less than 19, but not go above 19 if it's larger. So this should be handled, let me know if you see any problems. Thanks!

PR #502, opened 2021-06-17
working in progress - add symmetries function

fuhaoda:
This is a working in progress PR. It has added a function to mark the symmetric duplicative locations, and I also added a few tests. So far, it seems work fine. 

I am still trying to figure out how to add them to the root node search. 
One key issue is that some of the function in search.cpp is const. I was trying to figure out how we can minimize the changes of the search. cpp.


cpp/game/board.cpp
```diff
@@ -2639,3 +2639,57 @@ bool Board::isAdjacentToPlaHead(Player pla, Loc loc, Loc plaHead) const {
   );
   return false;
 }
+
+vector<bool> Board::markSymmetricDuplicativeLoc() {
+  vector<bool> isLocSymDuplicative(MAX_ARR_SIZE, false);
+  //this function is the same as in the nninputs.cpp getSymLoc function
```
lightvector, line 2645: No, please do not duplicate the code. This entire function should also move to be within the same SymmetryHelpers namespace, and use the pre-existing getSymLoc function.

Any time you are are duplicating the entire implementation of an existing function, you should reconsider and try to find an approach that avoids that.

lightvector: And yes, the symmetries code living in nninputs.h right now is a little weird. It should probably be moved to a separate .h and .cpp file. You are welcome to leave it alone for now, for someone else to improve later, or to fix it now in this PR, but please don't make it even worse. :)

cpp/game/board.cpp
```diff
@@ -2639,3 +2639,57 @@ bool Board::isAdjacentToPlaHead(Player pla, Loc loc, Loc plaHead) const {
...
+    if(flipX) { x = x_size - x - 1; }
+    if(flipY) { y = y_size - y - 1; }
+
+    if(transpose)
+      std::swap(x,y);
+    return Location::getLoc(x,y,transpose ? y_size : x_size);
+  };
+
+  vector<int> symTypes;
+  symTypes.reserve(7);
```
lightvector, line 2659: Here and elsewhere: do not hardcode 7.
Use the existing NUM_SYMMETRY_COMBINATIONS constant in nninputs.h

cpp/game/board.cpp
```diff
@@ -2639,3 +2639,57 @@ bool Board::isAdjacentToPlaHead(Player pla, Loc loc, Loc plaHead) const {
...
+
+  for(int symmetry = 1; symmetry <=7; symmetry++){
+    bool isBoardSym = true;
```
lightvector, line 2662: Add a separate function that takes a board and a symmetry and checks whether the board is symmetric under that symmetry?

cpp/game/board.cpp
```diff
@@ -2639,3 +2639,57 @@ bool Board::isAdjacentToPlaHead(Player pla, Loc loc, Loc plaHead) const {
...
+        if (colors[loc] != colors[Location::getLoc(x, y, x_size)]){
+          isBoardSym = false;
+          break;
+        };
+      }
+      if(!isBoardSym) break;
+    }
+
+    if (isBoardSym){
+      isLocSymDuplicative[0] = true; //use the first location to indicate if the board has at least one symmetric type.
```
lightvector, line 2676: If you find yourself having to write a comment explaining something because it is confusing or complicated, there needs to be a good reason why you are making it confusing or complicated, otherwise it is not good coding style.

Why not just have a separate boolean variable with an appropriate name?

cpp/game/board.cpp
```diff
@@ -2639,3 +2639,57 @@ bool Board::isAdjacentToPlaHead(Player pla, Loc loc, Loc plaHead) const {
...
+      symTypes.push_back(symmetry);
+    }
+  }
+
+  if(isLocSymDuplicative[0]){
```
lightvector, line 2681: Why do you need this if statement? This code looks correct even when symTypes is an empty vector. If the zero case needs no special handling, then just let it proceed naturally?

cpp/game/board.cpp
```diff
@@ -2639,3 +2639,57 @@ bool Board::isAdjacentToPlaHead(Player pla, Loc loc, Loc plaHead) const {
...
+    for (const auto symType: symTypes) {
```
lightvector, line 2682: The coding style in other repos may vary, but for this repo the normal style is to avoid using the "auto" keyword where possible. Generally it is only used for highly complex or templated types (container iterators, lambdas, etc).

lightvector (commented):
I know this is not supposed to be a finished and polished PR, but I'll go ahead and make some comments on good coding style.

lightvector:
> I am still trying to figure out how to add them to the root node search.
> One key issue is that some of the function in search.cpp is const. I was trying to figure out how we can minimize the changes of the search. cpp.

You should not need to change whether any of the functions are const. If there is, it probably is a sign that you are using the wrong approach?
`beginSearch should` compute a mask of the unique symmetric moves, which can be stored in a field of the Search class. Any time in the rest of the search you need to know the unique symmetric moves, you can just refer to the same precomputed field.

cpp/game/board.cpp
fuhaoda: For the above two comments, originally, I was thinking to use nninputs.h/cpp function. If I do that, it seems that I will break another rule. In your cpp [code review](https://github.com/lightvector/KataGo/tree/master/cpp), you design the codes in a very nice order based on the dependence, core -> game -> neuralnet -> search. That is the reason I didn't use the codes in the neuralnet. I was thinking  that it might be nice to move the getSymLoc function into board.h/cpp. But I am not sure if there are other parts of the codes using getSymLoc.

fuhaoda:
"beginSearch should compute a mask of the unique symmetric moves, which can be stored in a field of the Search class. Any time in the rest of the search you need to know the unique symmetric moves, you can just refer to the same precomputed field." I was thinking about the same. Great!

cpp/game/board.cpp
lightvector: See the comment above - ALL of this code should be moved into the SymmetryHelpers namespace. 

The solution isn't to duplicate code. The solution is to move the code to the correct place.

lightvector: So don't write it in board.h and board.cpp in the first place. You're writing symmetry-related code. Put it with the other symmetry-related code! :)

fuhaoda: Just followed this suggestion and changes are made  with a new commit.

cpp/game/board.cpp
fuhaoda: agreed and it has been addressed.

cpp/game/board.cpp
fuhaoda: changed.

fuhaoda:
This is a duplicate pull request. Rebased to the master as handle_symmetry_rebase

PR #519, opened 2021-07-05
add match options for small board

fuhaoda:
As we discussed last week, here are some settings loaded into `PlaySettings PlaySettings::loadForMatch(ConfigParser& cfg)`. So people can add additional parameter to increase the small board match opening.

### DIFF
--- a/cpp/program/playsettings.cpp
+++ b/cpp/program/playsettings.cpp
@@ -25,6 +25,10 @@ PlaySettings PlaySettings::loadForMatch(ConfigParser& cfg) {
   playSettings.resignThreshold = cfg.getDouble("resignThreshold",-1.0,0.0); //Threshold on [-1,1], regardless of winLossUtilityFactor
   playSettings.resignConsecTurns = cfg.getInt("resignConsecTurns",1,100);
   playSettings.compensateKomiVisits = cfg.contains("compensateKomiVisits") ? cfg.getInt("compensateKomiVisits",1,10000) : 100;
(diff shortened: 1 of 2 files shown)


cpp/program/playsettings.cpp
```diff
@@ -25,6 +25,10 @@ PlaySettings PlaySettings::loadForMatch(ConfigParser& cfg) {
   playSettings.resignThreshold = cfg.getDouble("resignThreshold",-1.0,0.0); //Threshold on [-1,1], regardless of winLossUtilityFactor
   playSettings.resignConsecTurns = cfg.getInt("resignConsecTurns",1,100);
   playSettings.compensateKomiVisits = cfg.contains("compensateKomiVisits") ? cfg.getInt("compensateKomiVisits",1,10000) : 100;
+  playSettings.policyInitAreaProp = cfg.contains("policyInitAreaProp") ? cfg.getDouble("policyInitAreaProp",0.0,1.0) : 0.04;
```
lightvector, line 28: This should not default to 0.04, to avoid changes for people who don't specify these parameters.

lightvector (changes requested):
One detail that needs to be fixed. Please be careful and apply your judgment and don't *just* trust me when I suggest copying some lines, I can make mistakes too.

cpp/program/playsettings.cpp
fuhaoda: I just made the default value to 0. In addition, we make the compensateAfterPolicyInitProb as 1 to make sure that the game is fair when `policyInitAreaProp` is set to a value but `compensateAfterPolicyInitProb` is not.

lightvector: Great!

PR #525, opened 2021-07-11
fix cmake compile issue when using mingw compiler under Windwos

fuhaoda:
I just made the line 338 in Cmake as
```
if(NOT (MSVC OR MINGW))
```
Thank you for fixing this.


lightvector:
Thanks to you too, for testing!

PR #533, opened 2021-07-29
add friendlyMaxVisitsFactor setting

bailsman:
Makes KataGo keep pace with the opponent by adjusting max visits
to how fast the opponent is playing.

Useful if you're running a ranked bot on a go server and want to be
friendly if the opponent is also playing fast, without sacrificing
playing strength if the opponent makes full use of the time control.

Using max visits makes the bot play naturally, responding quickly
if the move is expected and taking more time if it's unexpected.

(Of course, Katago's idea of what is obvious may differ from a human.)

### DIFF
--- a/cpp/command/gtp.cpp
+++ b/cpp/command/gtp.cpp
@@ -348,6 +348,10 @@ struct GTPEngine {
 
   double genmoveTimeSum;
 
+  ClockTimer gameTimer;
+  double gameGenmoveTimeSum;
+  double friendlyMaxVisitsFactor;
+
   GTPEngine(
     const string& modelFile, SearchParams initialParams, Rules initialRules,
     bool assumeMultiBlackHandicap, bool prevtEncore,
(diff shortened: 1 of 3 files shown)


lightvector:
Got around to taking a look at this - neat idea, yeah.

One question - since this change is fundamentally about not taking too much more time than the opponent, why is it in terms of visits? It seems like that makes it potentially harder to configure, since different hardware may have very different speeds of computation. Shouldn't it be adjusting the time to search instead of the visits? For merge into master and release, I'd be more have a parameter that was time-based, if the intent is to control time usage.

You may also be interested in:
https://github.com/lightvector/KataGo/blob/master/cpp/search/searchparams.h#L111

There are a bunch of time control parameters here to play with. Although they're not featured prominently in the example GTP config, the code for parsing them is there, and you can add the parameters to a config and parse them as you wish.
https://github.com/lightvector/KataGo/blob/master/cpp/program/setup.cpp#L606

There hasn't been a lot of testing, but you can do things like make KataGo play faster on "obvious" moves, play faster when it anticipated the current move and searched it a lot on the previous turn (even under a time-based setting not just a visit-based setting), play faster in the opening and endgame but slower on the midgame, etc. In conjunction with "don't use more total time than opponent" like you tried to implement in this feature, maybe enough testing could produce a really natural time usage pattern?

PR #534, opened 2021-08-05
handle errors/execptions when summarize sgfs files

fuhaoda:
passed test on the damaged sgfs files

### DIFF
--- a/python/summarize_sgfs.py
+++ b/python/summarize_sgfs.py
@@ -173,8 +173,12 @@ class GameResultSummary:
 
   def _add_a_single_sgf_string(self, sgf_string):
     """add a single game in a sgf string save the results in self.results."""
-    game = sgf.Sgf_game.from_bytes(sgf_string)
-    winner = game.get_winner()
+    try:
+      game = sgf.Sgf_game.from_bytes(sgf_string)
+      winner = game.get_winner()
+    except:
+      print ("A sgf string is damaged, and its record has been skipped!")
+      return
     pla_black = game.get_player_name('b')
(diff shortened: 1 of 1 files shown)


python/summarize_sgfs.py
```diff
@@ -173,8 +173,12 @@ def _add_one_sgf_file_to_result(self, sgf_file_name):
 
   def _add_a_single_sgf_string(self, sgf_string):
     """add a single game in a sgf string save the results in self.results."""
-    game = sgf.Sgf_game.from_bytes(sgf_string)
-    winner = game.get_winner()
+    try:
+      game = sgf.Sgf_game.from_bytes(sgf_string)
+      winner = game.get_winner()
+    except:
+      print ("A sgf string is damaged, and its record has been skipped!")
```
lightvector, line 180: This is not ideal, you should almost never try to catch all exceptions in python, since it will even catch the user pressing ctrl-c or trying to terminate the program. Instead, catch only the specific exception(s) that from_bytes will actually raise.

Additionally, this error message could be a bit more helpful. Ideally _add_a_single_sgf_string should be augmented to have a "debug_source" argument, which contains the file name and/or line number, so that the print message can display more helpful info to the user about trying to find where the problem is.

lightvector (changes requested):

python/summarize_sgfs.py
fuhaoda: I just updated this based on the comment to be more specific on the error as value error. In addition, I also made the print out error as colored message so that people can see more clear on where the error from.

PR #549, opened 2021-09-19
TensorRT backend

hyln9:
Hello!

This PR adds a fully-featured TensorRT backend to KataGo. By exploiting low-level APIs of TensorRT, it has the same user experience as other backends while being faster under the same precision and batch size requirements, especially when "requireMaxBoardSize" is turned on. Moreover, it supports caching during performance tuning and thus initialization time is drastically reduced afterwards.

The implementation has been thoroughly tested on a Linux machine with NVIDIA GPU. It should also work on a Windows one, although not tested.


hyln9:
Some tiny changes were accidentally missed during initial upload. Sorry for the forced push.

hyln9:
Some tiny changes were accidentally missed during initial upload. Sorry for the forced push.

hyln9:
Some tiny changes were accidentally missed during initial upload. Sorry for the forced push.

hyln9:
Some tiny changes were accidentally missed during initial upload. Sorry for the forced push.

hyln9:
Sorry for those forced pushes. Some tiny changes were accidentally missed during initial upload.

petgo3:
Just to document: I compiled Windows version with CMake/vs2019, running now on 2080ti without problem. Thx a lot

cpp/neuralnet/trtbackend.cpp
```diff
@@ -0,0 +1,1441 @@
...
+    }
+
+    vector<unique_ptr<float[]>> extraWeights;
+    auto modelParser = make_unique<ModelParser>();
+    auto model = modelParser->parse(move(network), loadedModel, ctx->nnXLen, ctx->nnYLen, requireExactNNLen);
+
+    bool saveTimingCache;
+    string timingCacheFile;
+    unique_ptr<ITimingCache> timingCache;
+    {
```
lightvector, line 891: All of this timing cache creation - is it potentially inaccurate if this runs at the same time as the GPU is under heavy usage by other parts of the code?

For OpenCL we have this issue, so when it performs tuning of the OpenCL parameters, there is code here: https://github.com/lightvector/KataGo/blob/master/cpp/command/contribute.cpp#L591 that performs tuning for *all* neural net sizes at startup.

Otherwise, we would end up performing some tuning whenever a new size is encountered in the middle of contribute, simultaneously with other threads using the GPU for other games. 

Should we do the same here?

lightvector (commented):
Took a first pass at reviewing and testing this, and making sure everything else works with it! Left some comments and questions for now, will take another pass later.

Also, I saw from the discord thread that this benefits from requireExactNNLen/requireMaxBoardSize a lot more than other backends? I'll experiment a little myself, but unfortunately one of the most compute-intensive applications (contribute) isn't going to be able to set this to true. Hopefully it's still a decent backend in that case?

cpp/CMakeLists.txt
```diff
@@ -151,6 +151,12 @@ if(USE_BACKEND STREQUAL "CUDA")
         "
         )
   endif()
+elseif(USE_BACKEND STREQUAL "TENSORRT")
+  message(STATUS "-DUSE_BACKEND=TENSORRT, using TensorRT backend.")
+  set(NEURALNET_BACKEND_SOURCES
+    neuralnet/trtbackend.cpp
+    )
+  set(CUDA_USE_STATIC_CUDA_RUNTIME OFF CACHE INTERNAL "")
```
lightvector, line 159: One more question that would help in testing and merging this - do you remember why you needed to add this line? For me, it's compiling okay without this line, and having this CUDA_USE_STATIC_CUDA_RUNTIME value set in the cache makes me unable to compile the CUDA version *after* I switch my cmake from TensorRT back to CUDA. Because I think after building with TensorRT, now this value is still in force even after going back to other backends because of setting it to cache.

hyln9:
Thanks for your detailed review! I'm on vacation these days and will start working on this soon.

Dreamkeeper66666:
Hi! Would you consider making it compatible with lower version of tensorrt, since `ITimingCache` is a new API in tensorrt 8. Is there any other way of implementing this? Thanks! @hyln9

hyln9:
I believe the CMakeLists.txt can also be rewritten in a much clearer way if we shift version requirement of CMake from 3.10.2 to 3.17 due to the introduction of CUDA language support and FindCUDAToolkit.

hyln9:
As for performance, it's still generally faster on RTX GPUs without requireExactNNLen/requireMaxBoardSize.

hyln9:
> Also, I saw from the discord thread that this benefits from requireExactNNLen/requireMaxBoardSize a lot more than other backends? I'll experiment a little myself, but unfortunately one of the most compute-intensive applications (contribute) isn't going to be able to set this to true. Hopefully it's still a decent backend in that case?

As for performance, it's still generally faster on RTX GPUs without requireExactNNLen/requireMaxBoardSize.

lightvector:
I'll reiterate that on the whole, I don't really understand what the best practices are for building and deploying precompiled exes involving CUDA and static/dynamic linking, and I'm happy if you know better, if you want to even just implement it to do it the best way, I'm happy to go along with it. However since what is currently there works for me, whereas what is in this PR (sometimes) doesn't work, I will need help with the build error causing it not to work, and the CMakeFile should also probably be fixed so that backends are independent and do not affect each other - suggestions are appreciated! :)

lightvector:
Hmmm it looks like I ran into this myself earlier and left this in the cmakelists:
```
  # On g++ it seems like we need to explicitly link threads as well.
  # It seems sometimes this is implied by other options automatically like when we enable CUDA, but we get link errors
  # if we don't explicitly require threads it when attempting to build without CUDA
  if(NOT USE_BACKEND STREQUAL "CUDA")
    find_package (Threads REQUIRED)
    target_link_libraries(katago Threads::Threads)
  endif()
```

Build configuration I always feel is such a mess and I still don't understand why the simple tweak of CUDA_USE_STATIC_CUDA_RUNTIME or not suddenly makes it necessary to explicitly link threads or not.

> I believe the CMakeLists.txt can also be rewritten in a much clearer way if we shift version requirement of CMake from 3.10.2 to 3.17 due to the introduction of CUDA language support and FindCUDAToolkit.

I'd been resisting this a little since "sudo apt install cmake" gives you 3.10.2 on Ubuntu 18.04, but if you think this will allow for making all of this build stuff less of a mess, we can definitely try it.

hyln9:
> > > Okay, fourth self-reply in same thread: I still don't really feel like I understand CUDA static vs dynamic linking tradeoffs, but I went ahead and tried a cmake cleanup as @hyln9 you suggested converting everything to static and bumping the required cmake version. Let me know what you think! Especially if you see me doing something stupid or if you see further things to clean up: [450303f](https://github.com/lightvector/KataGo/commit/450303f233829db0dece5464074142aa11299a5b)
> > > Also, I tested TensorRT backend on my own machine and it looks amazing, basically produces all the same results (modulo the tiny floating point changes that always happen when you change backends or even just change CUDA version, basically all my tests "pass"), and even on a V100 on a linux cloud machine is giving me something like a 20% to 30% speedup in one benchmark of a 60-block net, with enough threads. Thanks again for all the work setting this up!
> > > Assuming this handles the cmake stuff, it looks like the main remaining things are:
> > > 
> > > * Whether something needs to be done about the timing cache generation (this comment [TensorRT backend #549 (comment)](https://github.com/lightvector/KataGo/pull/549#discussion_r720912833))
> > > * Whether to import anything from [Desmond tensorrt #556](https://github.com/lightvector/KataGo/pull/556) (and is there some "core dump" crash bug to be fixed?),
> > > * And for me to also make sure I can build release executables for Windows as well as some other people have reported success building on Windows.
> > 
> > 
> > @lightvector Wow! I've incorporated changes from your branch and made a few more tweaks:
> > 
> > 1. FindCUDA is replaced with FindCUDAToolkit due to deprecation.
> > 2. Timing cache generation now performs on NULL stream so there are no more overlapped GPU activities.
> > 
> > I've also successfully compiled a Windows executable, yet it's not tested.
> > As for [Desmond tensorrt #556](https://github.com/lightvector/KataGo/pull/556), there is nothing to import in my view, but YMMV. And I cannot reproduce the "core dump" crash bug, maybe `-model` was missing? @inisis
> 
> image
> 
> 1. as for core dump, I use command "./katago genconfig" -model kata.bin.gz
> 2. I don't think serialize TimingCache is of good practice, I tested your code for many times, even TimingCache was generated, the engine creating time can be up to several minutes, but for my implementation, the engine creating time can take serveral minutes as yours, but once the engine cache is created, the new engine can be created in serveral seconds with a very small variance of time.

@inisis Serializing timing caches is one of the official way to reduce engine build time and works pretty well for many testers. Could you please turn on warning for TensorRT and paste its output here? Just replace kERROR with kWARNING at the following position.
https://github.com/lightvector/KataGo/blob/4fc7748c4bb7140efb407176d29d3d8774e0014a/cpp/neuralnet/trtbackend.cpp#L43

As for the crash, AddressSanitizer detected nothing on my code thus a stack trace is required. [This](https://stackoverflow.com/questions/5745215/getting-stacktrace-from-core-dump) may serve as an instruction.

inisis:
> > > > Okay, fourth self-reply in same thread: I still don't really feel like I understand CUDA static vs dynamic linking tradeoffs, but I went ahead and tried a cmake cleanup as @hyln9 you suggested converting everything to static and bumping the required cmake version. Let me know what you think! Especially if you see me doing something stupid or if you see further things to clean up: [450303f](https://github.com/lightvector/KataGo/commit/450303f233829db0dece5464074142aa11299a5b)
> > > > Also, I tested TensorRT backend on my own machine and it looks amazing, basically produces all the same results (modulo the tiny floating point changes that always happen when you change backends or even just change CUDA version, basically all my tests "pass"), and even on a V100 on a linux cloud machine is giving me something like a 20% to 30% speedup in one benchmark of a 60-block net, with enough threads. Thanks again for all the work setting this up!
> > > > Assuming this handles the cmake stuff, it looks like the main remaining things are:
> > > > 
> > > > * Whether something needs to be done about the timing cache generation (this comment [TensorRT backend #549 (comment)](https://github.com/lightvector/KataGo/pull/549#discussion_r720912833))
> > > > * Whether to import anything from [Desmond tensorrt #556](https://github.com/lightvector/KataGo/pull/556) (and is there some "core dump" crash bug to be fixed?),
> > > > * And for me to also make sure I can build release executables for Windows as well as some other people have reported success building on Windows.
> > > 
> > > 
> > > @lightvector Wow! I've incorporated changes from your branch and made a few more tweaks:
> > > 
> > > 1. FindCUDA is replaced with FindCUDAToolkit due to deprecation.
> > > 2. Timing cache generation now performs on NULL stream so there are no more overlapped GPU activities.
> > > 
> > > I've also successfully compiled a Windows executable, yet it's not tested.
> > > As for [Desmond tensorrt #556](https://github.com/lightvector/KataGo/pull/556), there is nothing to import in my view, but YMMV. And I cannot reproduce the "core dump" crash bug, maybe `-model` was missing? @inisis
> > 
> > 
> > image
> > 
> > 1. as for core dump, I use command "./katago genconfig" -model kata.bin.gz
> > 2. I don't think serialize TimingCache is of good practice, I tested your code for many times, even TimingCache was generated, the engine creating time can be up to several minutes, but for my implementation, the engine creating time can take serveral minutes as yours, but once the engine cache is created, the new engine can be created in serveral seconds with a very small variance of time.
> 
> @inisis Serializing timing caches is one of the official way to reduce engine build time and works pretty well for many testers. Could you please turn on warning for TensorRT and paste its output here? Just replace kERROR with kWARNING at the following position.
> 
> https://github.com/lightvector/KataGo/blob/4fc7748c4bb7140efb407176d29d3d8774e0014a/cpp/neuralnet/trtbackend.cpp#L43
> 
> As for the crash, AddressSanitizer detected nothing on my code thus a stack trace is required. [This](https://stackoverflow.com/questions/5745215/getting-stacktrace-from-core-dump) may serve as an instruction.

1. I generated engine cache file on my RTX 3090 and tested on my A6000(same compute capability), the result is as follow:
53d117739e779e4c47e6f77d47c6eaf

2. I removed the engine cache and generate it on A6000, the benchmark result is as follow:
12b4a83912804d09c8f2487bf54742e

3. I use the engine file generated by RTX3090 on another RTX3090, the benchmark result is as follow:
image

by using engine file generated by same type of device, no warning will appear, and the performance will not decrease; by using engine file generated by same compute capability deivce, waning will appear, but performance will not decrease either.

hyln9:
@inisis Did you turn on warning for TensorRT with **this PR** instead of #556?

Btw, I've improved the logic in case there are invalid timing caches, which should fix the crash and the overlong initialization time.

hyln9:
> > @inisis Did you turn on warning for TensorRT with **this PR** instead of #556?
> > Btw, I've improved the logic in case there are invalid timing caches, which should fix the crash along with the overlong initialization time.
> 
> image
> 
> 1. tensorrt warning is turned on, there is no other warning except for "Try increasing the workspace size to 4194304 bytes to get better performance." but core dump exits.
> 2. I tested your branch, with timing cache, the engine generation can be always around half a minutes on my RTX2080TI, which is not bad.
> 
> Anyway, thanks for your code, I learned a lot. GOOD JOB! @hyln9

Sounds great! I can see the original crash has gone. Could you please provide a stack trace of the new crash?

hyln9:
@inisis Assuming other parts of KataGo is correct and AddressSanitizer is working, the last crash can only happen in `freeComputeHandle`, which is most likely caused by either misconfiguration (e.g. library mismatch) or a bug inside TensorRT that I cannot fix.

@lightvector Looks like all issues are resolved for now.

hyln9:
> Thanks. I still have one remaining question:
> 
> ```
>     // So that there are no concurrent kernel executions probably from other parts of code
>     config->setProfileStream(cudaStreamLegacy);
> ```
> 
>  * Why are we confident that this is a good way to guard against other things running and will result in accurate timings? Is it possible that this may result in even _more_ inaccurate timings if other things are running because now the timing will get messed up by the fact that kernels will be waiting on other kernels to finish, and that delay itself will affect the timings of the cache?
> * Don't we need to set the stream back at some point to the per-thread non-legacy stream? Otherwise this will be a persistent performance cost any time anyone tries to run KataGo match using multiple networks, or tries to contribute while having a rating match ongoing with self-play games, right?

- Kernels launched in NULL stream start _after_ the implicit synchronization so it's duration time is not affected.
- The NULL stream is used solely for profiling and all inference workloads are launched in the per-thread default stream.

hyln9:
> > Kernel launched in NULL stream starts after the implicit synchronization so its GPU time is accurate.
> 
> Just checking, are you confident about this? When you profile a kernel in the way that the TensorRT builder is likely to do internally, does it only report the time within the kernel itself, and explicitly does not include the time cost or overhead in invoking or scheduling the kernel, or any blocking time involved in synchronizing the kernel with other streams?

TensorRT builder measures kernel time with CUDA events. Theoretically, the only race condition here is that the inference thread calls `cudaLaunchKernel` between `cudaEventRecord` and `cudaLaunchKernel` commands from the profiling thread, however, the chances are extremely small since both commands are asynchronous while the two host threads usually wait on different commands which are mutually exclusive due to implicit synchronization brought by the NULL stream and can hardly send commands to streams at the same time.

lightvector:
Hehe, yes there can be multiple compute contexts. You can see this from the fact that `const LoadedModel* loadedModel` is an input to it, so the interface forces there to be more than one any time there are multiple networks at a time - such as if running a match between two different networks (`katago match`, `katago gatekeeper`), or when contributing games in parallel (`katago contribute`) when the network changes and some games are still using the old net haven't finished while some other games start using a new net. 

Wow, thanks for the quick fix!

Are you aware of any other difficulties or quirks in the TensorRT API besides this one? Are there any other gotchas you know about for 8.2?

For documentation, would you also be able to add comments to CMakeLists about *why* this version is required, and can you also take a pass over the implementation and add comments regarding any other careful design choices in the backend code or any remaining oddities in TensorRT that you've tried to work around in the code? At least in this case, it would have saved a lot of debugging time for me if during review I could see a comment about the odd logger behavior and the attempted reason for putting the logger in the context instead of having to rediscover it myself.

lightvector:
Aside from documenting any remaining API quirks and workarounds, are there any things you are aware of regarding thread safety that should be considered? Threaded use of a backend is not heavily-tested by current tests since it is a bit tricky to write reliable tests with threads.

Almost the only thing that is guaranteed is that each compute handle is owned and used by one thread, and that a compute context and loaded model will live at least as long as the compute handles for it.

Other than that, there can be multiple networks, multiple compute contexts, multiple threads per compute context and multiple compute handles per network, including multiple compute handles even for the same GPU, and any of them can be created or destroyed or used at any time while concurrently any other creation destroying or usage of any other handle or context is happening.

I think the TensorRT backend is compatible with this, is your understanding of thread safety in TensorRT in agreement with this as well?

hyln9:
Sorry for the forced push. I've updated the commit according to the comments. There's also another minor commit on improving documentation.

Creation and destruction of `ComputeHandle` is associated with cuBLAS loading and unloading. There is a known issue in TensorRT 8.2 that cuBLAS can probably throw an error of `CUBLAS_STATUS_EXECUTION_FAILED` when cuBLAS is loaded and unloaded multiple times but it will be resolved in a future CUDA 11.4 update.

As far as I know, there are no other thread safety problems in TensorRT backend.

PR #555, opened 2021-10-03
adopts defaultBoardSize for benchmark

y-ich:
Hi.

I am not sure the original code is intentional or not, benchmark command ignores defaultBoardSize in a config file.
It means that benchmark without -boardsize option is always executed by 19x19 weight.

This pull request is a change that benchmark reads defaultBoardSize.


lightvector:
Thank you for the change! If we're doing this though, we should also pay attention to when the board size is specified separately for X and Y too. I'll make that change.

PR #556, opened 2021-10-05
Desmond tensorrt

inisis:
Thanks hyln9 for providing the base code, I modified the code and the improvement is as follow:

1. reduce engine create time by saving tenosrrt engine to file, only need to create once, time reduced from many minutes to a few seconds; 

2. make it compatible with cuda backend config, users don't have to create cfg file fot tenosrrt backend.


lightvector:
Thanks, the way you implemented (2) seems like it will break any configs that used "deviceToUse", because rather than adding a new option, you replaced an old option. And it may introduce inconsistency and confusion, because it only hardcodes "cuda" for one of the trt backend parameters, but not for all of the other backend-specific parameters.

Is it guaranteed that TensorRT will list all devices in the same order as CUDA, and is it likely to be the case that every user will always want to use the same parameters for CUDA as for TensorRT? If so, then rather than hackily hardcode one parameter, we can consider just making them both use the same backend prefix "cuda" instead of "trt" for the purpose of the config.

Additionally, for (1), when saving the engine to a file, does this file including the full network weights? If so, I think your change may be buggy, because it may continue using the same cached file with the old network weights even when the user specifies a new neural net to use.

By the way, if you have knowledge about any of the questions I asked in the original PR: https://github.com/lightvector/KataGo/pull/549 - help would be greatly appreciated! Finalizing these and any other details and we should be ready to merge and release soon.

hyln9:
Those two implementation choices were actually intended.

1. Serialized TensorRT engine contains full network weights and there are multiple modes involved such as precision and mask, therefore, profiling timing cache instead of engine was serialized so as to avoid bugs and storage overhead.

2. As @lightvector said, some parameters are not preferably shared between CUDA backend and TensorRT backend. For example, after benchmark, someone has found that best `numSearchThreads` value is higher for TensorRT backend on RTX GPUs.

hyln9:
> > Those two implementation choices were actually intended.
> > 
> > 1. Serialized TensorRT engine contains full network weights and there are multiple modes involved such as precision and mask, therefore, profiling timing cache instead of engine was serialized so as to avoid bugs and storage overhead while preserving acceptable initialization time afterwards. My implementation can also handle incompatibilities between TensorRT and driver versions.
> > 2. As @lightvector said, some parameters are not preferably shared between CUDA backend and TensorRT backend. For example, after benchmark, someone has found that the best `numSearchThreads` value of TensorRT backend is higher than CUDA backend's on RTX GPUs.
> 
> Thanks for your comment.
> 
> 1. for different modes, for example fp16 or fp32, my code will generate two engine files with different suffix and my implementation currently will not explicitly tell users the imcompatibility between different versions of tensorrt, but users can delete the older version tensorrt engien file and regenerate one, maybe I can add tensorrt version to the generated engine files. My standing is "engine only needs to be created once, load time reduced from many minutes to a few seconds;"
> 2. I tried to generate best cfg file by using [TensorRT backend #549](https://github.com/lightvector/KataGo/pull/549) binary file but got a core dump, if the cfg can be generated automatically, there is indeed no need to share the cfg file between cuda backend and tensorrt backend.

1. So there are multiple copies of weights from the same model, especially with combinations of modes. Unlike my implementation, the engine has to be built from scratch even if the network architecture does not change, apart from the storage overhead. There are APIs for refitting an engine but it's somehow cumbersome to use. I have done many experiments and found that timing cache is sufficient for reducing initialization time from minutes to seconds.

2. Thanks for your feedback and I'll have a look on that.

PR #590, opened 2021-12-17
More experimental fpu and weighting params

fuhaoda:


fuhaoda:
As we discussed, this is the fix

lightvector:
The parameters in this PR that were most promising have been extracted out and merged to master. Thanks for testing!

PR #591, opened 2021-12-26
Graphsearch

lightvector:
Experimental branch implementing graph search, as well as a lot of general search refactoring and cleanup.


sbbdms:
Seems the default value of graphSearchCatchUpLeakProb is 0.0. What's the proper gap if we want to select different value to assess? Is it 0.1 (0.0/0.1/0.2/0.3/0.4)?

lightvector:
Those are reasonable. It's a probability, so anything between 0 and 1 is reasonable. I expect too close to 1 is not good though.

petgo3:
Just feedback from windows trt version, after bugfix perfectly stabil for me!

PR #593, opened 2022-01-02
Add statistical data (surprise/entropy) to training output (for later analysis)

OmnipotentEntity:
This PR records the surprise and entropy for every turn into the training output (unused outputs C30-32), rather than making some sort of aggregate average.  If I recorded it to the wrong place, please let me know and I'll adjust it!

### DIFF
--- a/cpp/dataio/trainingwrite.cpp
+++ b/cpp/dataio/trainingwrite.cpp
@@ -24,6 +24,9 @@ SidePosition::SidePosition()
    pla(P_BLACK),
    unreducedNumVisits(),
    policyTarget(),
+   policySurprise(),
+   policyEntropy(),
+   searchEntropy(),
    whiteValueTargets(),
    targetWeight(),
    targetWeightUnrounded(),
@@ -36,6 +39,9 @@ SidePosition::SidePosition(const Board& b, const BoardHistory& h, Player p, int 
    pla(p),
    unreducedNumVisits(),
    policyTarget(),
+   policySurprise(),
+   policyEntropy(),
+   searchEntropy(),
    whiteValueTargets(),
    targetWeight(1.0f),
    targetWeightUnrounded(1.0f),
@@ -139,6 +145,13 @@ void FinishedGameData::printDebug(ostream& out) const {
     }
     out << endl;
   }
+  for (int i = 0; i < policySurpriseByTurn.size(); i++)
+    out << "policySurpriseByTurn " << i << " " << policySurpriseByTurn[i] << endl;
+  for (int i = 0; i < policyEntropyByTurn.size(); i++)
+    out << "policyEntropyByTurn " << i << " " << policyEntropyByTurn[i] << endl;
+  for (int i = 0; i < searchEntropyByTurn.size(); i++)
+    out << "searchEntropyByTurn " << i << " " << searchEntropyByTurn[i] << endl;
+
   for(int i = 0; i<whiteValueTargetsByTurn.size(); i++) {
     out << "whiteValueTargetsByTurn " << i << " ";
     out << whiteValueTargetsByTurn[i].win << " ";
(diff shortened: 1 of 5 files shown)


cpp/program/play.cpp
```diff
@@ -1520,7 +1522,13 @@ FinishedGameData* Play::runGame(
       gameData->nnRawStatsByTurn.push_back(computeNNRawStats(toMoveBot, board, hist, pla));
 
       gameData->targetWeightByTurn.push_back(limits.targetWeight);
-      policySurpriseByTurn.push_back(toMoveBot->getPolicySurprise());
+
+      double policySurprise = 0.0, policyEntropy = 0.0, searchEntropy = 0.0;
+      toMoveBot->getPolicySurpriseAndEntropy(policySurprise, searchEntropy, policyEntropy); // returns a bool, ignored, should always succeed in this case.
```
lightvector, line 1527: Maybe assert(the bool) instead of ignoring it?

cpp/dataio/trainingwrite.cpp
```diff
@@ -1015,6 +1031,7 @@ void TrainingDataWriter::writeGame(const FinishedGameData& data) {
             sp->unreducedNumVisits,
             &(sp->policyTarget),
             NULL,
+            0.0, 0.0, 0.0,
```
lightvector, line 1034: I guess we should add these values to the side position computation as well, since they do also form part of the training data. Check above to see where we basically do the same search we do on game positions for side positions, it should be straightforward to have sp->policySurprise, etc.

lightvector (commented):
Two inline comments. This seems great, it actually seems a lot cleaner than I initial way I had envisioned to do it, I hadn't though to make use of these unused outputs. I guess it's good to have a second opinion on the code. :)
 
My GPUs are busy running graph search testing (and also trying to transition to a cheaper setup than my current cloud setup), so if you have the ability to run this and measure some stats it would be great.

cpp/program/play.cpp
```diff
@@ -916,6 +916,15 @@ static void recordTreePositionsRec(
     SidePosition* sp = new SidePosition(board,hist,pla,numNeuralNetChangesSoFar);
     extractPolicyTarget(sp->policyTarget, toMoveBot, node, locsBuf, playSelectionValuesBuf);
     extractValueTargets(sp->whiteValueTargets, toMoveBot, node);
+
+    double policySurprise = 0.0, policyEntropy = 0.0, searchEntropy = 0.0;
+    bool success = toMoveBot->getPolicySurpriseAndEntropy(policySurprise, searchEntropy, policyEntropy);
```
OmnipotentEntity, line 921: You mentioned out of band that the changes to the side position calculation here to support finding policy surprise and entropy should be more complicated, but I cannot see how.  I suspect that it has something to do with the state of the `Search` object that I'm calling against.  Can you clarify?

OmnipotentEntity:
> My GPUs are busy running graph search testing (and also trying to transition to a cheaper setup than my current cloud setup), so if you have the ability to run this and measure some stats it would be great.

I absolutely do, but I'll probably have to drop off of regular contribution during that time.  Not a big deal certainly.

cpp/program/play.cpp
lightvector: The node passed in here to the other functions (like extractPolicyTarget) is not the root node of the search, whereas getPolicySurpriseAndEntropy doesn't take a node as an argument and only computes the value for the root node. That makes doing this correctly a bit messier. :)

cpp/search/searchresults.cpp
```diff
@@ -463,11 +463,15 @@ double Search::getPolicySurprise() const {
...
 }
 
-//Safe to call concurrently with search
 bool Search::getPolicySurpriseAndEntropy(double& surpriseRet, double& searchEntropyRet, double& policyEntropyRet) const {
-  if(rootNode == NULL)
+  return getPolicySurpriseAndEntropy(surpriseRet, searchEntropyRet, policyEntropyRet, rootNode);
+}
+
+//Safe to call concurrently with search
+bool Search::getPolicySurpriseAndEntropy(double& surpriseRet, double& searchEntropyRet, double& policyEntropyRet, const SearchNode* node) const {
```
OmnipotentEntity, line 471: I know that this isn't the *cleanest* way to implement this API, but is this acceptable?  Or will this result in a lot of multi-threading issues?

lightvector: I think this will work, thanks!

OmnipotentEntity:
Latest push to fix an oopsie, where I didn't actually insert the data required.  Derp derp.

lightvector:
Merged into master as 6320304e
Thanks!

PR #630, opened 2022-04-30
Config @include directive support with corresponding unittests

avtomaton:
- support @include directive in the config files
  - check for circular dependencies (not on the filesystem level though);
  - allow parameters overriding for existing parameters in included files;
  - option for overriding parameters in the same file (disabled by default);
- custom exception for config parsing (needed to distinguish between simple IOError and parsing errors);
- unit-tests for config parser;
- support for multiple config files from command line (for re-use common parts in different tasks without including them in specific configs);
- print final config content to all logging buffers.


AdamGleave:
Thanks for opening the PR @avtomaton.

It occurs to me we might want to actually put this change into practice with some of the default configs. For example, there's a lot of similarity between the configs in [training](https://github.com/lightvector/KataGo/tree/master/cpp/configs/training), especially `selfplay8a.cfg` and `selfplay8b.cfg`.

But I'll defer to @lightvector on whether he wants to see changes to the existing configs and, if so, whether he'd prefer it to be in this or a separate PR.

lightvector:
Just dropping in to say thanks for this PR and that it is not overlooked! I haven't had the chance in the last week or so to do a lot of coding work or review on KataGo, but I appreciate it and currently plan to make time to review and merge it next weekend. 

At a quick skim right now:
* It looks like there are many nice refactors of common code between the various commands that do config parsing or logging. These look great! My only comment is that in a few places this has resulted in passing bare bools or similar arguments into functions or constructors, this is the main thing I notice at a glance makes me a bit sad. Could we change it instead to declare const variables locally with the same name as the parameters and then pass those in instead? Given that C++ language doesn't yet have labeled argument passing, I've fallen back to this style in a few places elsewhere (e.g. https://github.com/lightvector/KataGo/blob/master/cpp/command/benchmark.cpp#L303). This of course doesn't guard against getting the variables in the wrong order, but it does make the code more locally self-documenting, and is no worse than the alternative in still providing compiler warnings/errors when  misordered variables happen to have different types or are accidentally omitted entirely resulting in the wrong the number of arguments.

* Regarding changes to the existing selfplay configs, I think I wouldn't mind seeing all the common parameters factored out for those. A separate PR would be great. For configs targeted at less-experienced users like gtp_example.cfg and analysis_example.cfg, I would definitely prefer to leave those as is however.

Thanks again!

AdamGleave:
Thanks for the review lightvector!

@avtomaton: the const bool change seems fine to me. For changing the
existing configs, I think we can probably just cherry-pick changes from our
internal branch for the selfplay configs to make a PR for this. I think for
our internal use case keeping all the configs factored out is probably
easier, although we may want to undo it in places in the future when merge
conflicts arise.

avtomaton:
Thanks for the review lightvector and Adam!

I updated the PR for passing non-default named parameters to the logger

lightvector:
This is now merged to master branch (rather than stable) as ec3d1e4b, with a minor merge conflict fix, followed by some commits making a few further cleanups, updates to the slower tests, and some nearby minor improvements to existing old code that I also noticed in the process.

Thanks!

PR #685, opened 2022-10-04
Add GTP kata-set-param maxVisits parameter

Emanuel-de-Jong:
Changing the MaxVisits on a running instance has a multitude of use cases. A few examples are:
- Different strength/calculation time for different stages of the game (e.g. quickly finishing in the endgame).
- AI vs AI games with different strengths per side.

The reason I made this is because I need it myself to not need multiple instances for a single game. (https://discord.com/channels/417022162348802048/583775968804732928/1026510646784839711)

I've build and tested it and it works without issues.

### DIFF
--- a/cpp/command/gtp.cpp
+++ b/cpp/command/gtp.cpp
@@ -571,6 +571,11 @@ struct GTPEngine {
     bot->setParams(params);
     bot->clearSearch();
   }
+  void setMaxVisits(int maxVisits) {
+    params.maxVisits = maxVisits;
+    bot->setParams(params);
+    bot->clearSearch();
+  }
 
   void updateDynamicPDA() {
     updateDynamicPDAHelper(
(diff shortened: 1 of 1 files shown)


Emanuel-de-Jong:
If maxVisits wasn't used before, I'd expect it to get enabled. But wouldn't disable the other limits on changing the maxVisits as the user/program probably wants the other limit(s) if they are set in the config. I feel this would be the most expected/intuitive behavior.
From what I understand, this is how it works already.

It might be an idea to make a parameter for maxPlayouts and maxTime as well. A value of 0 could give the same behavior as keeping the variable commented in the config. This makes the limits fully customizable.

lightvector:
Merged, along with a followup change that also adds maxPlayouts, maxTime, makes these parameters available in kata-get-params, and adds a quick mention of these in the GTP extensions documentation. Thanks!

PR #686, opened 2022-10-06
fixed an include path

y-ich:


ChinChangYang:
The include `../core/using.h` had troubled me for a few tens of minutes until I add `external` to the header search paths in Xcode project.

lightvector:
Thanks!

PR #701, opened 2022-11-11
Added missing gtp extension commands to documentation.

OmnipotentEntity:
It seems that when the `printsgf` command was added, it was not added to the documentation.  This PR is designed to rectify this.  While going through the rest of the exposed sgf commands I came across one other debugging command that isn't documented.  I have also added this; however, I'm not confident that this is necessarily the sort of command that ought to be documented, so let me know if you need me to remove that, (or if I got the description of the command wrong.)

### DIFF
--- a/cpp/command/gtp.cpp
+++ b/cpp/command/gtp.cpp
@@ -93,6 +93,7 @@ static const vector<string> knownCommands = {
 
   //Some debug commands
   "kata-debug-print-tc",
+  "debug_moves",
 
   //Stop any ongoing ponder or analyze
   "stop",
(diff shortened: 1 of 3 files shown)


lightvector:
Thanks for reporting! Happy to include printsgf.

I think the time control debug command would be reasonable to include in the documentation for better awareness, but it should be clearly described that it due to being a debug command, although users can feel free to use it, the existence, usage, or output format of that command is not promised to remain the same in future versions and may change without any warning.

The same goes for "debug_moves", which is actually missing from list_commands and probably should be added to it for completeness.

OmnipotentEntity:
I've added a few of the other undocumented debugging commands and put a warning on them.  I also added debug_moves to the known commands list.

lightvector:
Thanks!

PR #730, opened 2023-01-08
TensorRT backend v2

hyln9:
Hello!

This PR is a major update to the TensorRT backend which includes the following notable changes:

- Fix global pooling implementation for mish activation introduced in model version 11. This is already presented on the PyTorch side but missing in the TensorRT backend as of 1.12.0.
- Use explicit batch mode as TensorRT is dropping support for implicit batch. This requires a bunch of dynamic shape related changes since KataGo produces variable batch sizes. Variable board sizes are also possible now but not implemented due to nninterface limitation.
- Timing cache file are now specific to GPU type instead of physical device. Initialization time is reduced for machines with multiple identical GPUs (e.g. 8xA100) as cache is reused. Advanced users may further recompile and use engine plan caching instead which allows instant initialization at the cost of excessive disk space usage.
- TensorRT backend does not depend on cuDNN any more. It runs flawlessly with only TensorRT runtime library and compatible GPU drivers.
- Support for new GPU architectures such as Ada and Hopper from TensorRT 8.5 release.

Thanks for your hard work on the new neural net architecture!


hyln9:
Just realized that my recent changes have removed barriers that prevent usage of serialized TensorRT engines. I've added a commit that contains a implementation which can be activated by uncommenting `//#define CACHE_TENSORRT_PLAN`.

Once enabled, a cache hit allows instant engine loading at the cost of execessive disk space usage.

hyln9:
Just realized that my recent changes have removed barriers that prevent usage of serialized TensorRT engines. I've added a commit that contains a implementation which can be activated by uncommenting `//#define CACHE_TENSORRT_PLAN`.

Once enabled, a cache hit allows instant engine loading at the cost of excessive disk space usage.

hyln9:
Will rebase and make a clearer PR as the bug fix part is already merged.

PR #749, opened 2023-01-22
remove duplicate error messages for bad board size

rooklift:
`parseInteger()` in `analysis.cpp` generates an `{error}` object (for the GUI) when it fails; it also returns false, but when the main loop sees the false return value, it also generates an error, leading to 2 `{error}` objects being sent to the GUI.

This PR removes the duplication.

Note: not actually tested. I've no C++ compiler.


lightvector:
Thanks!

Issue #387, opened 2021-01-16
Compiling lasted master source with errors 


qiqiguaitm:
Using Ubuntu18.04  install CUDA 10.2 and cudnn7.6.5.32

cmake . -DUSE_BACKEND=CUDA
make with errors:
CMakeFiles/katago.dir/build.make:459: recipe for target 'CMakeFiles/katago.dir/neuralnet/cudahelpers.cu.o' failed
make[2]: *** [CMakeFiles/katago.dir/neuralnet/cudahelpers.cu.o] Error 1
make[2]: *** Waiting for unfinished jobs....
In file included from /root/explore/KataGo/cpp/neuralnet/cudabackend.cpp:12:0:
/root/explore/KataGo/cpp/neuralnet/../external/half-2.1.0/include/half.hpp: In function ‘unsigned int half_float::detail::float2half_impl(float, std::true_type)’:
/root/explore/KataGo/cpp/neuralnet/../external/half-2.1.0/include/half.hpp:912:42: warning: there are no arguments to ‘_mm_set_ss’ that depend on a template parameter, so a declaration of ‘_mm_set_ss’ must be available [-fpermissive]
    return _mm_cvtsi128_si32(_mm_cvtps_ph(_mm_set_ss(value),
                                          ^~~~~~~~~~
/root/explore/KataGo/cpp/neuralnet/../external/half-2.1.0/include/half.hpp:913:34: error: ‘_MM_FROUND_TO_NEAREST_INT’ was not declared in this scope
     (R==std::round_to_nearest) ? _MM_FROUND_TO_NEAREST_INT :
                                  ^~~~~~~~~~~~~~~~~~~~~~~~~
/root/explore/KataGo/cpp/neuralnet/../external/half-2.1.0/include/half.hpp:914:35: error: ‘_MM_FROUND_TO_ZERO’ was not declared in this scope
     (R==std::round_toward_zero) ? _MM_FROUND_TO_ZERO :
                                   ^~~~~~~~~~~~~~~~~~
/root/explore/KataGo/cpp/neuralnet/../external/half-2.1.0/include/half.hpp:914:35: note: suggested alternative: ‘__FD_ZERO’
     (R==std::round_toward_zero) ? _MM_FROUND_TO_ZERO :
                                   ^~~~~~~~~~~~~~~~~~
                                   __FD_ZERO
[...]

PR #755, opened 2023-02-04
Upgrade half library to 2.2

espadrine:
Compiling with
`cmake . -DUSE_BACKEND=CUDA -DCMAKE_CXX_FLAGS='-march=native'` on a CPU with F16C intrinsics yielded the same errors as issue #387:

> error: ‘_mm_set_ss’ was not declared in this scope

The cause is that _mm_set_ss is defined in a header imported from immintrin.h, which is included in
cpp/external/half-2.1.0/include/half.hpp:272.

However, it is only included if the preprocessor macro HALF_ENABLE_F16C_INTRINSICS is set, but in turn, that is only set in cpp/external/half-2.1.0/include/half.hpp:283, a few lines too late.

This issue seems fixed in half-2.2.0, thus we upgrade to it.

I compiled and ran the new code in GTP against Sabaki and played a game without issue.


lightvector:
Thanks! Manually rebased and merged into master as 3dd5999f along with some followup changes 2f7a95f7 a5cd339f to reduce the whitespace diff and to reapply local changes to make compiler warnings happy.

PR #758, opened 2023-02-10
maybe a bug when parsing rules

hzyhhzy:


lightvector:
Merged as 5cada8a9 along with a small test, thanks!

PR #777, opened 2023-03-22
Fix segmentation fault when using the match option "extraPairs"

OmnipotentEntity:
The option "extraPairs" does not initialize the bots found in the pairs, because they are excluded by the "includeBots" command.  This duplicates some parsing logic, which might be better consolidated elsewhere, but rather than making that call, I just copy pasted the code.

### DIFF
--- a/cpp/command/match.cpp
+++ b/cpp/command/match.cpp
@@ -80,6 +80,37 @@ int MainCmds::match(const vector<string>& args) {
     }
   }
 
+  vector<bool> extraPairsBot(numBots);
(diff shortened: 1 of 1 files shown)


lightvector:
Thanks for finding and debugging the cause of this problem! I fixed it in what should be an overhaul of the match logic, moving the existing ugly pairing generation logic around flow so fixes like this no longer require code duplication: https://github.com/lightvector/KataGo/commit/462df452ffd0197cdadaf080a56ea66ebcf37766

PR #787, opened 2023-05-16
Remove superfluous buffer

nerai:
It is cleared anyway, so it can just as well be put on the stack.

### DIFF
--- a/cpp/search/search.cpp
+++ b/cpp/search/search.cpp
@@ -1081,8 +1081,7 @@ bool Search::runSinglePlayout(SearchThread& thread, double upperBoundVisitsLeft)
   //Store this value, used for futile-visit pruning this thread's root children selections.
   thread.upperBoundVisitsLeft = upperBoundVisitsLeft;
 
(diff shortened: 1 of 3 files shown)


lightvector:
Thanks. Is there a reason to do 
```
bool posesWithChildBuf[NNPos::MAX_NN_POLICY_SIZE] = { false, }; // Initialize the first value explicitly to false, and all other values by default to false
```
instead of this?
```
bool posesWithChildBuf[NNPos::MAX_NN_POLICY_SIZE] = {}; // Initialize all to false
```

nerai:
> Thanks. Is there a reason to do
> 
> ```
> bool posesWithChildBuf[NNPos::MAX_NN_POLICY_SIZE] = { false, }; // Initialize the first value explicitly to false, and all other values by default to false
> ```
> 
> instead of this?
> 
> ```
> bool posesWithChildBuf[NNPos::MAX_NN_POLICY_SIZE] = {}; // Initialize all to false
> ```

@lightvector I am sorry, I missed your response. No, in fact I prefer your version.

PR #803, opened 2023-06-11
update to avoid trt API Usage Error

inisis:
for trt version 8.6, there will be an error like: TensorRT backend: 3: [runtime.cpp::~Runtime::346] Error Code 3: API Usage Error (Parameter check failed at: runtime/rt/runtime.cpp::~Runtime::346, condition: mEngineCounter.use_count() == 1. Destroying a runtime before destroying deserialized engines created by the runtime leads to undefined behavior.

this pr helps fix it

### DIFF
--- a/cpp/neuralnet/trtbackend.cpp
+++ b/cpp/neuralnet/trtbackend.cpp
@@ -923,6 +923,7 @@ struct ComputeHandle {
   TRTLogger trtLogger;
   map<string, void*> buffers;
   unique_ptr<ICudaEngine> engine;
+  unique_ptr<IRuntime> runtime;
   unique_ptr<IExecutionContext> exec;
 
   ComputeHandle(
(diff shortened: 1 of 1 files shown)


lightvector:
Thanks! I notice a similar fix has been also submitted by @hyln9 in this PR: https://github.com/lightvector/KataGo/pull/805/files - but there is a difference, in that this one PR a few extra calls to .reset() for various things at the end. Is this an important change that should be added to the other PR?

hyln9:
The unique_ptr automatically deletes the object it points to when it goes out of scope. So, calling reset() manually is not necessary and is taken care of by the unique_ptr's destructor.

inisis:
the truth is that the sequence of releasing the object matters, or the tensorrt will give warnings.

hyln9:
> the truth is that the sequence of releasing the object matters, or the tensorrt will give warnings. @hyln9 @lightvector

And that's why I put `unique_ptr<IRuntime> runtime;` before `unique_ptr<ICudaEngine> engine;` in my patch. Declaration order does matter here.

For more information on this topic, see the "Destruction sequence" section in the [C++ Reference](https://en.cppreference.com/w/cpp/language/destructor).

PR #804, opened 2023-06-14
Typedef the progress state of a SearchNode

nerai:
To clarify the meaning of the variables.


lightvector:
Great idea, thanks! There were some additional spots missed in this PR, so I added them in this followup commit 0fc0ffb8, which also moves the typedef out of the struct and duplicates it once to avoid a circular dependency between search.h and searchnode.h which both want to use that type.

PR #806, opened 2023-06-16
add plan cache to cmake opts

kinfkong:
so that in cmake, we can use:
-DUSE_CACHE_TENSORRT_PLAN=1
or
-DUSE_CACHE_TENSORRT_PLAN=0
to turn on/off the trt plan cache


lightvector:
Thanks!

PR #826, opened 2023-08-03
Replace Board reference with just board size

nerai:
Avoid duplicate work,
Simplify calculations,
Reduce dependencies.

### DIFF
--- a/cpp/game/board.cpp
+++ b/cpp/game/board.cpp
@@ -242,7 +242,7 @@ void Board::setSimpleKoLoc(Loc loc) {
 }
 
 
-double Board::averageSideLength() const {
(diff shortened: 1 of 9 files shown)


nerai:
Good idea, I'll push an update.

lightvector:
Don't mean to nitpick, but it should be sqrtBoardArea instead of sqrtSideLength, right? Like, when the board is 10 x 16, this value will be equal to sqrt(160), rather than equal to the square root of either of the side lengths, sqrt(10) or sqrt(16).

nerai:
Oops, my bad. Sorry! Fixed.

lightvector:
Looks great, thanks!

Issue #879, opened 2024-01-01
katago 1.14.0 TRT plan cache boots significantly slower than 1.13.2


kinfkong:
Here is the testing environments for comparing the booting time for 1.13.2 and 1.14.0 when hitting the plan cache (That means the plan cache files already exist).

For 1.13.2: using TensorRT 8.5.2
For 1.14.0: using TensorRT 8.6.1
The  loading weight is: 18b

In a 5 cards of RTX3080 machine, it takes **40 seconds** for 1.14.0 to boot to GTP ready, while 1.13.2 just need **17 seconds**
In a 8 cards of RTX4070 machine, it takes **63 seconds** for 1.14.0 to boot to GTP ready, while 1.13.2 just need **26 seconds**.

I also try with different weights and different machines, 1.14.0 is generally boots much slower than 1.13.2

Have you @lightvector or @hyln9 observed this? Thanks!

hyln9:
I do not currently have access to these GPUs.

Could you please try commenting out [this line](https://github.com/lightvector/KataGo/blob/c6de1bbda837a0717eaeca46102f7326ed0da0d4/cpp/neuralnet/trtbackend.cpp#L999) and see if it makes a difference?

Otherwise it might be a bug in TensorRT itself.

kinfkong:
[1.13-trt8.5-benchmark.txt](https://github.com/lightvector/KataGo/files/13805262/1.13-trt8.5-benchmark.txt)
[1.14-trt8.6-benchmark.txt](https://github.com/lightvector/KataGo/files/13805263/1.14-trt8.6-benchmark.txt)

@hyln9 
1. Commenting out that line does not help. Commenting that line out or not ,does not affect the boot time, they are the same.
2. I attached the benchmark log for katago1.13+trt8.5 and katago1.14+trt8.6. benchmark log shows more details, and you can check the timestamps in the log to check if there are any hints? thanks.
3. The benchmark also shows there is a about 10% performance drop in 1.14.

Maybe the trt 8.6 itself causes that: 1). cost more time to start katago. 2). katago benchmark is worse in trt 8.6.

cc: @lightvector

kinfkong:
[1.14-trt8.5-benchmark.txt](https://github.com/lightvector/KataGo/files/13805445/1.14-trt8.5-benchmark.txt)
I slightly modify the code to make 1.14 to comapitable with trt 8.5. Here is the benchmark log. 
Everything seems fine running katago 1.14 in trt 8.5

[...]

PR #880, opened 2024-01-01
Make Katago 1.14 compatible with TRT 8.5

kinfkong:
for issue #879


kinfkong:
close this issue, because a better pr #882 fixed this.

PR #882, opened 2024-01-05
Restore TensorRT 8.5 support

### LINKED ISSUE #879: katago 1.14.0 TRT plan cache boots significantly slower than 1.13.2

hyln9:
Closes #879


lightvector:
Thanks!

PR #891, opened 2024-01-27
README: fix grep usage in macOS instructions

jaysephjw:
The current command has a slightly broken pattern, and fails to find `gtp_example.cfg`.


lightvector:
Thank you!

PR #910, opened 2024-03-09
remove repetitive words

gcmutator:


OmnipotentEntity:
I do not see the the problem. 😂

lightvector:
Thanks for pointing out the typos! I pushed the appropriate fixes as 659e3e9a. In particular, I did *not* fix the typos you identified in tclap, because that is an external library and I don't want to make custom modifications unless absolutely necessary.

Issue #913, opened 2024-03-18
Unable to compile - error: invalid conversion from ‘float**’ to ‘void**’


toomasr:
I had to compile KataGo again and realised that I cannot get it to compile. I've tried CUDA 11.7 and 12.4 and also many different g++ versions but still no luck. I also tried to compile different tags so I presume I'm doing something wrong as I'm unable to get any of those compiled and the errors are the same. Any ideas?

```
-- The C compiler identification is GNU 11.4.0
-- The CXX compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Building 'katago' executable for GTP engine and other tools.
-- -DUSE_BACKEND=CUDA, using CUDA backend.
-- The CUDA compiler identification is NVIDIA 12.4.99
-- Detecting CUDA compiler ABI info
-- Detecting CUDA compiler ABI info - done
-- Check for working CUDA compiler: /usr/local/cuda/bin/nvcc - skipped
-- Detecting CUDA compile features
-- Detecting CUDA compile features - done
-- Including Git revision in the compiled executable, specify -DNO_GIT_REVISION=1 to disable
-- Found Git: /usr/bin/git (found version "2.40.1")
-- Found CUDAToolkit: /usr/local/cuda/include (found version "12.4.99")
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found version "1.2.13")
-- Setting up build for GNU or Clang.
-- Enabling GNU-specific build options.
-- Configuring done (2.0s)
-- Generating done (0.0s)
-- Build files have been written to: /home/toomasr/projects/KataGo/cpp
[  0%] Generating program/gitinfoupdated.h
[  1%] Building CXX object CMakeFiles/katago.dir/core/global.cpp.o
[  2%] Building CXX object CMakeFiles/katago.dir/core/base64.cpp.o
[...]

PR #923, opened 2024-04-05
CUDA 12.3 and 12.4 support (#913)

tlyx:
cast to void** in cudaMalloc calls.

### DIFF
--- a/cpp/neuralnet/cudabackend.cpp
+++ b/cpp/neuralnet/cudabackend.cpp
@@ -2096,22 +2096,22 @@ struct Buffers {
     inputGlobalBufBytesFloat = m.numInputGlobalChannels * batchFloatBytes;
     inputGlobalBufBytes = m.numInputGlobalChannels * batchBytes;
 
-    CUDA_ERR("Buffers",cudaMalloc(&inputBufFloat, inputBufBytesFloat));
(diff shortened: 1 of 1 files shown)


tlyx:
The template version of function cudaMalloc for C++ is actually defined in cuda_device_runtime_api.h. And it is included in c++ source file by cuda_runtime_api.h. 

From CUDA 12.3,  you should define some macros (Line 152-154 of cuda_runtime_api.h): 
#if defined(__CUDACC_RDC__)  || defined(__CUDACC_EWP__) || !defined(__CUDACC_RTC__)
#include "cuda_device_runtime_api.h"
#endif /* defined(__CUDACC_RDC__) || defined(__CUDACC_EWP__) || !defined(__CUDACC_RTC__) */

The line "#if defined" and "#endif" are newly added.  So the template version of cudaMalloc is not defined when we compile c++ code.

Cast to void** in cudaMalloc calls  can solve the problem and keep compatiblility with previous CUDA.

lightvector:
Apologies, I haven't had the bandwidth to install CUDA 12.3 or 12.4 myself and test things out, so I put off responding to this. Do you still think this is a reasonable change? If so, you can just leave it open and I will get to it probably around the time I assemble the next release and put together all the various pieces and test them.

If you are closing this because there is a better way, or because it's unneeded/redundant somehow, then no worries.

PR #965, opened 2024-07-26
Fix missing cudaMalloc error

tlyx:
Fix missing cudaMalloc error in #935


tlyx:
Already done in v1.15.2.

PR #983, opened 2024-09-13
Fix to arguments to GameState() in play.py

awswan:
I noticed a minor bug in play.py. At some point the arguments to the constructor for GameState were changed to take ruleset as well as board size, but a couple of places it's used in play.py are still calling it with only board size. I just added Tromp-Taylor rules as argument, since that's what's done earlier on in the file.


lightvector:
Was fixed as part of other changes, thanks for the report.

PR #992, opened 2024-10-08
Eval cache rebased

lightvector:
Experimental branch with an option (`useEvalCache=true`) that caches evals of nodes greater than a configurable number of visits (`evalCacheMinVisits`) and remembers them permanently for the duration of that running instance of KataGo, to bias subsequent searches of the same positions to converge to the correct answer faster.

Right now there is no way to clear this cache other than restarting KataGo, and there is no way to save the cache to disk in order to reuse it on a subsequent run. These are things that could be added if this were made into a proper feature.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -227,6 +227,7 @@ add_executable(katago
   search/localpattern.cpp
   search/searchnodetable.cpp
   search/subtreevaluebiastable.cpp
+  search/evalcache.cpp
   search/patternbonustable.cpp
   search/analysisdata.cpp
   search/reportedsearchvalues.cpp
--- a/cpp/command/misc.cpp
+++ b/cpp/command/misc.cpp
@@ -1358,6 +1358,7 @@ int MainCmds::dataminesgfs(const vector<string>& args) {
   double gameModeFastThreshold;
   bool flipIfPassOrWFirst;
   bool allowGameOver;
+  bool manualHintOnly;
   double trainingWeight;
 
   int minRank;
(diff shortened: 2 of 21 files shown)


kaorahi:
This feature suggests a new style of whole-game analysis, at least for less powerful PCs. In the following example, doing "2sec scans three times" seems better than doing a "single 6sec scan".

Before getting into the main result, I'd like to mention the issue I'm focusing on. Here's black's winrate chart from a single 2sec scan for each move (please ignore the vertical bars).



The red and green segments show bad and good moves --- in other words, moves that lower or raise the player's winrate. Note that no green segments would appear if the AI were perfect. In this chart, you can see alternating red and green segments along a monotonically decreasing part of black's winrate. This makes it hard to tell which moves are bad and how bad they are. Generally, small green segments are OK, but long ones are problematic unless the move truly surpasses the AI.

The next charts show the results from the "single 6sec scan" (left), "2sec scans three times" (center), and "2sec scans three times without eval cache" (right).





The center chart highlights the bad moves more clearly than the left chart. While the right chart has a similar tendency, the center chart appears slightly clearer. Of course, KataGo was restarted for each chart, and GUI's analysis cache was disabled during this experiment.

lightvector:
@kaorahi - Very cool. Is this phenomenon consistent across multiple different games? It would be interesting to know if one could also find other ways to try to objectively measure evaluation quality and compare to see if making use of the eval cache in this way improves the quality by those other measurements.

kaorahi:
Sorry, this is just a quick initial report, and I don't have time for serious experiments right now. I'll report more results when I get them.

PR #1000, opened 2024-11-13
Loading the next npz file while yielding the current file when training

hzyhhzy:
It will reduce read and write bottlenecks and accelerate small net training


lightvector:
Thank you for the suggestion! Implemented in a very similar way but avoiding mutating the input array and also with adjusted log messages, in https://github.com/lightvector/KataGo/commit/a968f0114934c14223299aa936caeee8ff9944be

PR #1022, opened 2025-02-07
Fix typo in GraphSearch.md

GoodCoder666:
Minor fix, replacing `it's` with the correct form `its`. Thanks for the great docs!


lightvector:
Thanks.

PR #1042, opened 2025-03-31
Allow playing cleanup moves with rootPruneUselessMoves.

d0sboots:
This allows analysis to rate/recommend these moves, even if one side is passing. It also means cleanup will usually (but not *always*) proceed through this path when playing via GTP, with the cleanup hack providing backup only.

This shouldn't unduly affect the length of training games, since the number of cleanup moves that can be played is strictly bounded, and most training games are terminated as soon as all territory is pass-alive anyway.

### DIFF
--- a/cpp/search/searchhelpers.cpp
+++ b/cpp/search/searchhelpers.cpp
@@ -313,7 +313,8 @@ bool Search::isAllowedRootMove(Loc moveLoc) const {
   //A bad situation that can happen that unnecessarily prolongs training games is where one player
   //repeatedly passes and the other side repeatedly fills the opponent's space and/or suicides over and over.
   //To mitigate some of this and save computation, we make it so that at the root, if the last four moves by the opponent
-  //were passes, we will never play a move in either player's pass-alive area. In theory this could prune
+  //were passes, we will never play a move in either player's pass-alive area. We make an exception for playing in
+  //own territory next to opponent's stones, since that is needed for cleanup. In theory this could prune
   //a good move in situations like https://senseis.xmp.net/?1EyeFlaw, but this should be extraordinarly rare,
   if(searchParams.rootPruneUselessMoves &&
      rootHistory.moveHistory.size() > 0 &&
@@ -330,7 +331,9 @@ bool Search::isAllowedRootMove(Loc moveLoc) const {
        rootHistory.moveHistory[lastIdx-2].pla == opp &&
        rootHistory.moveHistory[lastIdx-4].pla == opp &&
        rootHistory.moveHistory[lastIdx-6].pla == opp &&
-       (rootSafeArea[moveLoc] == opp || rootSafeArea[moveLoc] == rootPla))
+       (rootSafeArea[moveLoc] == opp ||
+         (rootSafeArea[moveLoc] == rootPla && !rootBoard.isAdjacentToPla(moveLoc, opp))
+       ))
       return false;
   }
 


lightvector:
Thanks for the suggestion! I think I would like to address this in a way that more directly encodes this in the game rules object, so that it won't globally change this pruning, because changing this pruning will cause a sudden change in the data distribution for official self-play training, which it would be better to avoid. It won't make the next release, but I'll look into it as something to do for a later release.

d0sboots:
Is pruneUselessMoves used for selfplay training? If so then I could see that, since training isn't applying the gtp hacks. 
Ultimately very little will change for training here though, since the score and final winner remain invariant.

lightvector:
Yep, the winner and score will stay the same, but it will produce a sudden change in the distribution of the policy target in positions where the pruning would have occurred and assign weight to moves that had exactly zero probability before, and it will lengthen self-play games a bit and be less compute-efficient, which are the two main possible effects I'm concerned with.

d0sboots:
The first one *ought* to be a positive change, because these are the only sensible moves to play in the cleanup phase, so if they were being pruned before then the policy was learning to play strictly bad moves, which could explain a few things...

As to the second, according to comments only 2% of games are continued (not immediately ended) once all territory is pass-alive. So the amount of extra training length is very strictly bounded; in practice I would be surprised if it exceeded .05% (not all training games play by area rules, and the cleanup phase isn't as long as the main phase)

lightvector:
> The first one ought to be a positive change, because these are the only sensible moves to play in the cleanup phase, so if they were being pruned before then the policy was learning to play strictly bad moves, which could explain a few things...

Careful! In rules where pass-alive area is automatically awarded, playing additional moves in that area is *not* sensible, rather, it's making moves outside that area (converting non-pass-alive areas into pass-alive) that are the sensible ones. (I know you disagree on the value of training in such rules to begin with, and I think I agree with you on a lot of that, and might have done KataGo differently in an alternate world, but taking as a given that those actually are the rules that the training is learning under, the pruning is removing only non-sensible moves here, rather than pruning any sensible moves).

> I would be surprised if it exceeded .05% (not all training games play by area rules, and the cleanup phase isn't as long as the main phase)
Regardless of area vs territory ultimately the same thing has to happen - you have to go from a board that would end roughly where humans would end it via two passes, to a board where dead groups are physically captured and removed. The same moves have to happen either way (all the same liberty and eye filling moves of the same dead groups), so adjusting the pruning in this way should on average add a similar *absolute* number of additional moves during that process regardless of whether they're in a main phase or a cleanup phase, namely any additional moves spent filling liberties of the dead stones in pass-alive area following four passes by opponents. So I think the rules of area vs territory shouldn't matter much.

The amount that this pruning saves overall is not entirely trivial, mainly due to chopping short certain pathological behaviors that contribute a lot of moves to occasional games, and by training the policy itself to be efficient since the pruning also affects the target. But yes, the amount of lengthening by making just your proposed change is probably small.

d0sboots:
> Careful! In rules where pass-alive area is automatically awarded even when opponent's dead stones are in it, playing additional moves in that area is _not_ sensible, rather, it's making moves outside that area (converting non-pass-alive areas into pass-alive) that are the sensible ones. (I know you disagree on the value of training in such rules to begin with, and I think I agree with you on a lot of that, and might have done KataGo differently in an alternate world, but taking as a given that those actually are the rules that the training is learning under, the pruning is removing only non-sensible moves here, rather than pruning any sensible moves).
Yes. According to the rules that training operates under, *all* moves made after all territory is pass-alive are pointless/silly, since they can't affect the score. We still ask the policy network about such moves when we are playing in situations where cleanup is required, and hope for a sensible result, which is a rather odd thing to hope for.

(I'm not even criticizing this decision; endgame score is the obvious and most regular thing to train for/predict, and trying to tweak that leads to irregularities and oddities.)

I've been focusing specifically on when all the territory is pass alive; when this is not the case there are dame and/or sekis. In the first case this code isn't invoked because both players are taking turns to fill, so it's not relevant here. The second case is the more general case where instead of all moves having equal value, some are distinctly worse. This does not change the overall analysis, though.

So my meaning of "sensible" was not "within the rules it is training on" (all moves are equal in that sense) but rather sensible in the context of "within any situation we would actually want the network's opinion about this position."

PR #1078, opened 2025-06-30
Fix integer precision warnings in MetalProcess::processScoreValues

ChinChangYang:
This pull request addresses integer precision warnings in the MetalProcess::processScoreValues function within the metalbackend.cpp file. The local variable numScoreValueChannels has been updated from int to size_t to prevent implicit conversions and resolve compiler warnings.


lightvector:
Thanks.

PR #1107, opened 2025-09-21
Fix undefined variable in get_input_feature function

loker404:
Fix undefined variable in get_input_feature function


lightvector:
Thanks.

PR #1122, opened 2025-11-08
Fix endless recursions (Win) in threadsafequeue.h

### LINKED ISSUE #1119: `ThreadTest` doesn't work on Windows on the latest master (freezing)
Also, compiler reports the following, probably related: `'ThreadSafeContainer<int>::waitPush<int>': recursive on all control paths, function will cause runtime stack overflow`

KvanTTT:
fix #1119

### DIFF
--- a/cpp/core/threadsafequeue.h
+++ b/cpp/core/threadsafequeue.h
@@ -107,7 +107,8 @@ class ThreadSafeContainer
   inline std::enable_if_t<std::is_copy_constructible<U>::value, bool>
   waitPush(const T& elt)
   {
-    return waitPush(T(elt));
+    T copy(elt);
+    return waitPush(std::move(copy));
   }
 
   // Push an element without blocking, but can exceed maxSize of the queue.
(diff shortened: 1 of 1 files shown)


lightvector:
Interesting that different compilers are a bit inconsistent here. Tends to be more awkward for me to test on windows so thanks for debugging this!

ChinChangYang:
I’ve been using continuous integration (CI) in my personal projects. It’s great for spotting problems across different platforms, but it can be a bit of a distraction and make it hard to concentrate on building new features. I definitely suggest using CI for big projects with lots of developers who can help fix bugs on different platforms. But for smaller projects, I’m not sure if CI really adds much value or if it might even dampen the developer’s enthusiasm for creating.

KvanTTT:
> But for smaller projects, I’m not sure if CI really adds much value or if it might even dampen the developer’s enthusiasm for creating.

For me KataGo is already quite a big project that makes introducing of continious integration reasonable. Currently it has many forks that rely on the latest master. I understand your argument about developer's enthusiasm, however, nothing stops to experiment with whatever you want in a separated branch but not on master.

PR #1127, opened 2025-11-13
Introduce GitHub CI with basic built on Windows, Linux, macOS (OpenCL) and tests

KvanTTT:
I'm introducing minimal functionality for now, but it can be extended later to support other backends (CUDA, TensorRT) and more tests. Maybe a single iteration of `synchronous_loop.sh` also should be executed to make sure it's robust. Also, we should balance between execution performance and coverage (compilation of C++ is slow).

But even that covers a lot of stuff. For instance, [the recent regression](https://github.com/lightvector/KataGo/issues/1119).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,8 +43,9 @@ jobs:
         make -j$(nproc)
 
     - name: Run tests
+      working-directory: cpp
       run: |
-        cpp/katago runtests
+        ./katago runtests
 
     - name: Upload artifact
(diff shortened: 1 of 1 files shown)


KvanTTT:
@lightvector is it possible to merge it? It's small, but quite important pull request.

KvanTTT:
I've also improved publishing of Windows artifacts, now it contains all necessary `.dll` to run and it's self-contained.

PR #1136, opened 2025-12-22
Introduce default parameters for some appropriate functions and merge some signatures

KvanTTT:


lightvector:
Some parts of this were pulled in as https://github.com/lightvector/KataGo/commit/3ef68bfb5c128aa85859d71414a8bbf38bef0015 and also I fixed a major bug where min() was used for floating point types instead of lowest(). For floats, min is a tiny positive float, not a large negative float.

The ones that involve parameter rearrangements I'm not as interested in.

KvanTTT:
Thanks.

PR #1148, opened 2026-01-05
Improve Metal backend with hybrid CPU+GPU+ANE inference for macOS

ChinChangYang:
## Summary

This PR adds a Metal neural network backend for macOS that leverages Apple's full compute stack—CPU, GPU, and Apple Neural Engine (ANE) simultaneously—through a per-thread multiplexer architecture.

## Key Features

- **Per-Thread GPU/ANE Multiplexer:** Each NN server thread is dedicated to either GPU (MPSGraph) or ANE (CoreML). Multiple threads run in parallel, saturating all compute units without intra-batch splitting overhead.
- **Native Model Conversion:** Uses the [katagocoreml](https://github.com/chinchangyang/katagocoreml-cpp) C++ library for on-the-fly conversion from KataGo's .bin.gz format to CoreML .mlpackage (no Python dependency).
- **Configurable Precision:** FP16 (default) enables both GPU and ANE paths; FP32 uses GPU-only MPSGraph, bypassing CoreML conversion entirely.
- **Dynamic Batch Size:** Supports runtime batch sizes from 1 to maxBatchSize without model recompilation.
- **Flexible Dispatch Configuration:** Users configure compute unit assignment per thread via `metalDeviceToUseThread<N>` (0 = GPU, 100 = ANE), allowing GPU-only, ANE-only, or mux mode.

## Architecture

The backend uses a **multiplexer** design where each NN server thread owns exactly one compute handle:

- **GPU threads** (`metalDeviceToUseThread<N>=0`): Use MPSGraph for Metal GPU inference
- **ANE threads** (`metalDeviceToUseThread<N>=100`): Use CoreML for CPU+ANE inference

Example mux configuration (recommended for best throughput):
```
numNNServerThreadsPerModel = 4
metalDeviceToUseThread0 = 0    # GPU
metalDeviceToUseThread1 = 0    # GPU
metalDeviceToUseThread2 = 100  # ANE
metalDeviceToUseThread3 = 100  # ANE
```

## Performance

Benchmarked on Apple M3 Max with mux mode (2 GPU + 2 ANE threads), 2048 visits, `-half-batch-size`:

**b18c384nbt** (18 blocks, 384 channels):

| Threads | visits/s | nnEvals/s | avgBatchSize |
|---------|----------|-----------|--------------|
| 8       | 735.68   | 570.97    | 1.46         |
| 16      | 817.86   | 629.23    | 2.70         |
| 32      | 810.71   | 660.98    | 5.02         |

**b28c512nbt** (28 blocks, 512 channels):

| Threads | visits/s | nnEvals/s | avgBatchSize |
|---------|----------|-----------|--------------|
| 8       | 325.57   | 256.51    | 1.47         |
| 16      | 335.40   | 263.98    | 2.61         |
| 32      | 319.61   | 255.57    | 4.91         |

Per-thread workload distribution (b18c384nbt, 32 threads):
- GPU threads: ~24,200 rows processed
- ANE threads: ~24,572 rows processed

## Files Changed

| File | Description |
|------|-------------|
| `cpp/neuralnet/metalbackend.cpp` | C++ interface: model conversion, batch processing, data marshaling |
| `cpp/neuralnet/metalbackend.h` | Backend header with compute handle and layer descriptors |
| `cpp/neuralnet/metalbackend.swift` | CoreML model loading and per-thread compute handle orchestration |
| `cpp/neuralnet/metallayers.swift` | MPSGraph layer implementations for GPU inference path |
| `cpp/CMakeLists.txt` | Build configuration for Metal backend (Ninja + Swift) |
| `cpp/command/benchmark.cpp` | `-half-batch-size` option for benchmarking |
| `cpp/configs/gtp_example.cfg` | Metal backend dispatch configuration documentation |
| `cpp/configs/analysis_example.cfg` | Metal backend dispatch configuration documentation |
| `.github/workflows/build.yml` | CI job for Metal backend on macOS |
| `Compiling.md` | Build instructions for Metal backend |

## Build Requirements

- macOS 13.0+
- Xcode Command Line Tools with Swift 5.9+
- Ninja build system: `brew install ninja`
- katagocoreml library: `brew tap chinchangyang/katagocoreml-cpp && brew install katagocoreml`

```bash
cd cpp
cmake -G Ninja -DUSE_BACKEND=METAL -DBUILD_DISTRIBUTED=1
ninja
```

## Test Plan

- [x] `./katago runtests` passes
- [x] `./katago benchmark` verifies performance with mux mode (GPU+ANE)
- [x] GPU-only mode (`metalDeviceToUseThread0=0`) works correctly
- [x] ANE-only mode (`metalDeviceToUseThread0=100`) works correctly
- [x] FP16 mode: both GPU and ANE paths active
- [x] FP32 mode: GPU-only MPSGraph execution (CoreML bypassed)
- [x] Batch size 1: single compute handle works correctly
- [x] CI build succeeds on GitHub Actions


.github/workflows/build.yml
```diff
@@ -97,6 +97,56 @@ jobs:
...
+        run: |
+          ./katago runtests
+
+      - name: Upload artifact
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master'
+        uses: actions/upload-artifact@v4
+        with:
+          name: katago-macos-coreml
+          path: cpp/katago
+
```
KvanTTT: Is it possible to merge macOS build into a single configuration? As I see some build steps are overlapping.

cpp/neuralnet/metalbackend.swift
```diff
@@ -1,3032 +1,546 @@
...
+                        maskInput: maskInput,
+                        policy: policy,
+                        policyPass: policyPass,
+                        value: value,
+                        scoreValue: scoreValue,
+                        ownership: ownership
+                    )
+                } catch {
+                    printError("Metal backend: CoreML inference error: \(error)")
+                }
```
lightvector, line 126: Should this do some sort of harder failure? What happens if an error happens and this just continues onward - is there a possibility of uninitialized or bad outputs?

cpp/neuralnet/metalbackend.swift
lightvector, line 1: It seems like we have a new dependency now on https://github.com/ChinChangYang/coremltools/tree/katagocoremltools/katagocoreml which replicates a whole bunch of the detailed neural net logic that is also present via the mps graph path in metallayers.swift?

I'm a bit sad that there is now there is a dependency on a different repo and things are no longer self-contained and self-containedly testable and could be broken by future changes to the other repo? Especially because the code in this other looks like katago-specific architecture stuff, and if we ever change the katago architecture here it now requires nonlocal changes in a different repo. Is there a way to be more self-contained?

cpp/neuralnet/metalbackend.swift
```diff
@@ -1,3032 +1,546 @@
...
+    /// Copy MLMultiArray data to destination buffer, respecting strides.
+    /// Core ML may return non-contiguous arrays, especially for spatial outputs after GPU computation.
+    private func copyMultiArray(
+        _ array: MLMultiArray,
+        to dest: UnsafeMutablePointer<Float32>,
+        destOffset: Int
+    ) {
+        let shape = array.shape.map { $0.intValue }
+        let strides = array.strides.map { $0.intValue }
+        let ptr = array.dataPointer.assumingMemoryBound(to: Float32.self)
```
lightvector, line 218: Should there be some assertion or defensive protection about the type of the data in the tensor, rathre than assuming Float32?)

cpp/neuralnet/metalbackend.swift
```diff
@@ -3146,18 +632,21 @@ struct Model {
...
+            executionDescriptor: nil)
+
+        commandBuffer.commit()
+        commandBuffer.waitUntilCompleted()
 
-        assert(fetch[policyHead.policyTensor] != nil)
-        assert(fetch[policyHead.policyPassTensor] != nil)
-        assert(fetch[valueHead.valueTensor] != nil)
-        assert(fetch[valueHead.scoreValueTensor] != nil)
-        assert(fetch[valueHead.ownershipTensor] != nil)
```
lightvector, line 3159: Should these assertions still be present? Are they no longer needed, and what happens if in fact they would have fired but we just proceed onward silently? I presume that if the code is correct then they should in theory never fire, but it seems good to have them and perhaps even upgrade them to unconditional checks so that if we ever do have a bug we have a loud fail rather than silently proceeding on?

cpp/neuralnet/metalbackend.cpp
```diff
@@ -5,97 +5,167 @@
...
     case ACTIVATION_MISH_SCALE8:
-      testAssert(false); // Metal does not use scaled mish activations due to no fp16
-      return ActivationKind::identity(); // Placeholder for compilation
+      return ActivationKind::identity(); // Metal/CoreML does not use scaled mish
     case ACTIVATION_IDENTITY:
       return ActivationKind::identity();
     default:
-      testAssert(false);
-      return ActivationKind::identity(); // Placeholder for compilation
+      return ActivationKind::identity();
```
lightvector: Why does this default exist? Shouldn't it remain a testAssert(false) or better yet a throw StringError("some useful message") so that if we have a new activation kind implemented and not updated here, we fail loud and don't compute the wrong result?

cpp/neuralnet/metallayers.swift
```diff
@@ -0,0 +1,2914 @@
...
+
+    init(
+        graph: MPSGraph,
+        descriptor: SWMatMulLayerDesc,
+        sourceTensor: MPSGraphTensor
+    ) {
+        assert(
+            (sourceTensor.shape?.count == 4) || (sourceTensor.shape?[1] == descriptor.inChannels))
+        assert(
+            (sourceTensor.shape?.count == 2) || (sourceTensor.shape?[1] == descriptor.inChannels))
```
lightvector: What were these asserts meant to check? Is this equivalent to 

```
assert(
    (sourceTensor.shape?.count == 4) ||
    (sourceTensor.shape?.count == 2) ||
    (sourceTensor.shape?[1] == descriptor.inChannels))
```
And is that really the intended check? The expected state is that the tensor is 4-dimensional (with no restrictions on input channels), OR 2-dimensional  (with no restrictions on input channels), OR *any other dimensional* with a restriction that the index 1 matches the input channels?

cpp/neuralnet/metallayers.swift
```diff
@@ -0,0 +1,2914 @@
...
+// MARK: - Core Layers
+
+/// A class that represents a convolutional layer using MPSGraph
+class ConvLayer {
+    let resultTensor: MPSGraphTensor
+    let convDescriptor = MPSGraphConvolution2DOpDescriptor(
+        strideInX: 1,
+        strideInY: 1,
+        dilationRateInX: 1,
+        dilationRateInY: 1,
```
lightvector, line 492: These are hardcoded - if the katago model ever specifies a dilation that is not 1 (SWConvLayerDesc has dilationX and dilationY matching the katago common model desc from neuralnet/desc.h) this will be silently wrong?

(I think this was not new in this PR and was present before, but maybe it's still worth addressing and guarding against)

cpp/neuralnet/metalbackend.swift
```diff
@@ -1,3032 +1,546 @@
...
-/// Extension to MPSGraph to the mish activation function
-extension MPSGraph {
-    /// This function applies the Mish activation function on the input tensor `x`. The Mish function is defined as
-    /// x * tanh(Softplus(x)), where Softplus(x) is defined as log(1 + exp(min(x, 10.39))) if x < 10.39 and x otherwise.
-    /// When FP16 is later used, the threshold of softplus will need to be modified to 10.39, which is different from
-    /// the original 20. This is because exp(10.39) = 32532.666936 < 32767.0 < 65504.0, so the result of exp(10.39) can
-    /// be represented by float16. If the threshold of softplus is 20, the result of exp(20) is 485165195.40979004,
-    /// which is out of range of float16.
-    /// - Parameter tensor: The input tensor of mish activation function
-    /// - Returns: The output tensor of mish activation function
```
lightvector, line 79: The old code had this comment. 

* Does the new code support fp16?
* Is this addressed in the new code? A threshold is 20 is too large for fp16, and I don't think I see "10.39" in the new code.

lightvector (commented):
@ChinChangYang I finally got around to taking a look at this, sorry for the long delay in feedback, it was kind of daunting to look at a diff this big, and I still don't actually understand a lot of the diff that well.

I left some various comments and questions. Some of this review was AI-assisted, since I don't actually know swift or the macos system and APIs.

cpp/neuralnet/metalbackend.swift
ChinChangYang: This is a software limitation of C++ and Swift interoperability where C++ cannot catch errors from Swift. If an error happens, this just continues. The outputs are very likely to be bad.
It would be safer to throw a runtime error that doesn't need C++ to catch. Let me see how to implement this in the next revision.

cpp/neuralnet/metalbackend.swift
ChinChangYang: The new dependency should actually be this one: https://github.com/chinchangyang/katagocoreml-cpp. The main reason for this change is to enhance the user experience for using Core ML. This eliminates the need for users to figure out how to convert a network into a Core ML model and where or how to integrate a Core ML model into KataGo.

It’s possible to copy all the source code from `katagocoreml-cpp` to this repository, making it more self-contained. However, if you examine what `katagocoreml-cpp` compiles, you might reconsider if you think it’s too large.

As a conservative measure, I suggest that this PR be halted as an option for users who want Core ML acceleration.

cpp/neuralnet/metalbackend.swift
ChinChangYang: Assuming Float32 is similar to `(float32*)` in C++, it is essential to adhere to the output type, which is Float32.

cpp/neuralnet/metalbackend.swift
ChinChangYang: These assertions are valid during code-level inspection, but they’re removed from the source code for simplicity because they never trigger in actual use cases.  For some development reasons, these assertions are considered noise.

cpp/neuralnet/metalbackend.cpp
ChinChangYang: I agree with this.

cpp/neuralnet/metallayers.swift
ChinChangYang: Currently, the known network doesn’t have `dilation > 1`, so I don’t have a test case to verify that. However, it’s worth noting for future work.

cpp/neuralnet/metalbackend.swift
ChinChangYang: The new code doesn't support fp16. MPSGraph should always run in fp32 mode.

cpp/neuralnet/metallayers.swift
ChinChangYang: These assertions are designed to verify the dimensions of the source tensor. If an assertion fails, it indicates that the source tensor is incorrect, suggesting that the developer should debug the source tensor rather than the MatMul layer. These assertions proved incredibly useful when I implemented the Metal backend from scratch, so I would like to retain them.

cpp/neuralnet/metalbackend.cpp
ChinChangYang: Fixed.

cpp/neuralnet/metalbackend.swift
ChinChangYang: Fixed.

cpp/neuralnet/metallayers.swift
lightvector: I think you misunderstood my comment. My comment was saying that maybe these assertions are not strict enough? For example, to give a silly ridiculous example, a 10-dimensional tensor whose shape[1] happens to match the descriptor.inChannels would be accepted by both asserts here, even though a 10-dimensional tensor definitely shouldn't be present. 

My comment was asking - what *actually* are the shapes of tensors expected here, and should the assert be made stricter? Given the way the two asserts are equivalent to a single assert 
```
assert(
    (sourceTensor.shape?.count == 4) ||
    (sourceTensor.shape?.count == 2) ||
    (sourceTensor.shape?[1] == descriptor.inChannels))
```
it makes it look like maybe there's a typo here that makes the asserts less strict than intended.

cpp/neuralnet/metalbackend.cpp
lightvector: Thanks!

cpp/neuralnet/metallayers.swift
lightvector: Can we have an check that makes it fail loudly if dilation > 1? The current network doesn't have it, yes, but the current network file format allows specifying it, and if it were to be specified, the code seems like it would be silently wrong.

cpp/neuralnet/metalbackend.swift
lightvector: Thanks for the corrected link. What do you mean "I suggest that this PR be halted". What does "halted" mean?

ChinChangYang: I suggest that this PR be isolated in an option for users who want Core ML acceleration, so that this PR won't be merged into the master branch before resolving the dependency issue. I will find time to copy all the source code from `katagocoreml-cpp` to this repository, making it more self-contained.

cpp/neuralnet/metallayers.swift
ChinChangYang: I must admit that I misunderstood your comment.

The assertions were initially created to identify programming errors, but it appears that they have become obsolete or have been modified due to errors at some point.

Removed by b8bd04b6c08955b022e58d595ea4654f56f9f2fa.

cpp/neuralnet/metallayers.swift
ChinChangYang: Fixed by 6c3df7afd47cb3af06f42a1fd5841ec61936b3ac.

cpp/neuralnet/metalbackend.swift
ChinChangYang: Fixed by https://github.com/lightvector/KataGo/pull/1148/commits/2b8aa73faf327b4389c3d51c580d50a413047481, https://github.com/lightvector/KataGo/pull/1148/commits/32e7209168de32f5cd6839f844fb55ad27bb63e9, and https://github.com/lightvector/KataGo/pull/1148/commits/b0adf3e08e11c71fc5b2c594f491878a9ad18e22.

lightvector:
Merged! Finally got around to building and testing this a bit and it looks good. Thanks for all this work. Note that I made some followup changes in https://github.com/lightvector/KataGo/commit/8db4e41eba7d4c32b197cd8a09b6e769defd9243, mostly adding additional error checks, and also fixing a bug in the git hash checking when building using ninja.

PR #1157, opened 2026-02-10
Fix undefined variable and invalid escape sequence in play.py

changcheng967:
`play.py` has two bugs:

**Line 460**: `N` is undefined — causes `NameError` when the GTP `boardsize` command 
is called with an incompatible size. Replaced with `features.pos_len`.

**Line 451**: `'\d+'` is an invalid escape sequence. Python 3.12+ raises 
`SyntaxWarning` for this, and future versions will make it a hard error. 
Changed to `r'\d+'`.

### DIFF
--- a/python/play.py
+++ b/python/play.py
@@ -448,7 +448,7 @@ while True:
     if line == '':
         continue
     command = [s.lower() for s in line.split()]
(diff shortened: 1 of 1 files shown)


lightvector:
Thanks!

PR #1166, opened 2026-03-09
Make `Board` references constant where possible to improve clarity and avoid unnecessary copies

KvanTTT:
It looks like `NNEvaluator::evaluate` is a hot path and it would be nice to get rid of redundant copies. And it looks consistent because `BoardHistory` is also passed by const reference to `evaluate`.

### DIFF
--- a/cpp/book/book.cpp
+++ b/cpp/book/book.cpp
@@ -2758,7 +2758,7 @@ int64_t Book::exportToHtmlDir(
     if(hist.encorePhase > 0)
       return;
 
-    Board board = hist.getRecentBoard(0);
+    const Board& board = hist.getRecentBoard(0);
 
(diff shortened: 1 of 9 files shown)


lightvector:
Thanks for the change! I had to check things over to verify that the board (and the hist) actually is const in places, but I think this looks good.

KvanTTT:
Thanks for the merging!

Actually, I've discovered much more unnecessary copies (using CLang-Tidy). The things should be treated carefully, because in some places copy syntax is used intentionally and it's incorrectly to use const references. But they should be emphasized explicitly.

Also, I've found a lot of places where `push_back` can be replaced with `emplace_back` and other performance problems. I will create a meta-issue to track them and will prepare merge requests step by step.

Issue #1167, opened 2026-03-11
Meta-issue for performance improvements found by Clang-Tidy


KvanTTT:
* [performance-inefficient-vector-operation](https://clang.llvm.org/extra/clang-tidy/checks/performance/inefficient-vector-operation.html)
  * [x] `emplace_back` -> `push_back` on corresponding call sites. It looks trivial and safe.
  * [x] Use `reserve` to avoid unnecessary memory reallocations. It's also quite straightforward.
* [performance-unnecessary-copy-initialization](https://clang.llvm.org/extra/clang-tidy/checks/performance/unnecessary-copy-initialization.html)
  * [ ] Preliminary: [readability-non-const-parameter](https://clang.llvm.org/extra/clang-tidy/checks/readability/non-const-parameter.html)
  * [ ] Preliminary: [readability-make-member-function-const](https://clang.llvm.org/extra/clang-tidy/checks/readability/make-member-function-const.html)
  * [ ] Main step. I've found a lot of places with redundant copying, it could improve performance.

KvanTTT:
Fixed by https://github.com/lightvector/KataGo/commit/92ad583d4852ebaacd5f0398f052bdb537affe19

PR #1168, opened 2026-03-12
Fix performance inefficient vector operation

KvanTTT:
The fix is quite straight forward. If fixes the first two items from #1167


lightvector:
Thanks!

PR #1188, opened 2026-04-20
[AMD GPU] Add Windows & Linux ROCm support

Looong01:
All test passed, we can merge it to main branch! @lightvector

Bothe **Windows and Linux** Binary release has been published here: https://github.com/Looong01/KataGo-Multi-backends/releases

## Background
This PR summarizes all commits by `Looong01` on the `AMD_GPU` branch from `2025-07-28` to `2026-03-16` (`23` commits total: `18` non-merge + `5` merge), focused on introducing and refining ROCm backend support in KataGo, plus the new MIGraphX backend added on the `MIGraphX` branch.

## Key Changes — ROCm Backend
- Added core ROCm backend implementation and utility files: rocmbackend.cpp, `rocmhelpers.hip`, `rocmutils.*`, `rocmincludes.h`, `rocmerrorcheck.h`.
- Integrated `USE_ROCM_BACKEND` into startup and config flow (`setup/benchmark/gtpconfig`) for proper backend detection and config generation.
- Expanded ROCm build logic in CMakeLists.txt: HIP compiler setup, architecture/FP16 detection, ROCm package lookup, and fallback linking.
- Added Windows ROCm build support (`HIP_PATH/ROCM_PATH`, clang toolchain handling, Windows library search paths).
- Iteratively improved rocmbackend.cpp with Convlayer method updates, performance tuning, and bug fixes.
- Introduced and then removed experimental `rocmbackend_new.cpp` after merging validated changes into the main backend path.
- Updated docs and sample configs (README / Compiling / `cpp/configs/*`) with ROCm instructions and `rocmDeviceToUse*`, `rocmUseFP16` examples.
- Regularly merged upstream `lightvector:master` to reduce branch drift.

### Critical Bug Fix: ConvLayer accumulate (residual skip connections)
- **Root cause**: `miopenConvolutionForwardImmediate` does **not** support `alpha`/`beta` parameters (unlike cuDNN's `cudnnConvolutionForward`). The original code set `beta = accumulate ? 1.0 : 0.0` but this value was never passed to the MIOpen API, causing all residual skip connections to be silently dropped — the neural network output was effectively garbage.
- **Fix**: When `accumulate=true`, save the output buffer (trunk) to a pre-allocated `accumBuf` via `hipMemcpyAsync` (Device-to-Device), run convolution (which overwrites the output buffer), then add the saved residual back using a new `customCudaAddTensorsInplace` GPU kernel. All operations stay in VRAM with zero CPU-side data transfer.
- **New kernels added** in `rocmhelpers.hip` / `rocmhelpers.h`:
  - `customCudaAddTensorsInplace(float*, const float*, int)` 
  - `customCudaAddTensorsInplace(half*, const half*, int)`
- **Performance optimization**: `accumBuf` is pre-allocated once per `ConvLayer` at construction time (sized for `maxBatchSize`), avoiding per-inference `hipMalloc`/`hipFree` overhead.

### Secondary Fix: Algorithm enumeration buffer overflow
- `miopenConvolutionForwardGetSolutionCount` returns the available count by overwriting the output parameter. The original code used this count to size a fixed stack array `miopenConvSolution_t solutions[2*requestedAlgoCount]` which could overflow. Replaced with `std::vector<miopenConvSolution_t> solutions(availableAlgoCount)` for safe dynamic sizing.

### Windows ROCm Build — CMakeLists.txt Self-Configuration

Added full Windows ROCm build support directly into `CMakeLists.txt.

## Key Changes — MIGraphX Backend (New)

Added a complete MIGraphX graph-compiler backend (migraphxbackend.cpp, 1886 lines) as an alternative to the ROCm (MIOpen) backend. MIGraphX compiles the entire neural network into a single fused GPU program, leveraging AMD's graph-level optimizations (operator fusion, memory planning, kernel scheduling).

### Architecture
- **Graph-based inference**: The full model (trunk → policy/value/ownership heads) is constructed as a `migraphx::program` at load time using `MIGraphXGraphBuilder`, compiled once, then cached as `.mxr` files under `~/.katago/migraphxcache/`.
- **Multi-batch compilation**: MIGraphX does not support dynamic batch dimensions. Instead, multiple fixed-batch programs are pre-compiled at sizes `{4, 8, 16, 24, 32, 40, 64}` (capped by `maxBatchSize`). At inference time, `getBestBatchSize()` selects the smallest compiled size ≥ actual batch to minimize GPU waste.
- **Cache system**: Compiled programs are saved/loaded as `.mxr` files with naming format `migraphx_{modelName}_{sha256}_{H}x{W}_batch{N}_fp{0|1}_nhwc{0|1}_{exact|max}.mxr`. First launch compiles all batch sizes (slow); subsequent launches load from cache in seconds.

### Neural Network Components Implemented
| Component | Status |
|-----------|--------|
| Convolution (with padding/dilation/stride) | ✅ |
| BatchNorm (merged scale/bias via unsqueeze+multibroadcast) | ✅ |
| Residual Block | ✅ |
| Global Pooling Residual Block (dual-branch with gpoolToBiasMul) | ✅ |
| Nested Bottleneck Residual Block | ✅ |
| Global Pooling (3 features: mean, mean×scale, max/scale2) | ✅ |
| Policy Head (p1Conv + g1Conv→GPool→bias, p2Conv spatial, pass branch) | ✅ |
| Value Head (v1Conv→GPool→v2Mul→v3Mul) | ✅ |
| Score Value Head (shared v2→sv3Mul) | ✅ |
| Ownership Head (vOwnershipConv→flatten) | ✅ |
| Activations: ReLU, Mish, MishScale8 | ✅ |
| MatMul / FC layers (with optional bias) | ✅ |
| SGF Metadata Encoder | ❌ Disabled |

### FP16 Support
- Enabled by default (when `useFP16Mode` is `Auto` or `True`).
- Input parameters remain `float_type` on host; a `convert` op inside the graph handles float→half on GPU.
- All weights/computation use `half_type`; outputs are cast back to `float` via `static_cast<float>` in the `visit()` lambda.

### Build Integration
- New `USE_BACKEND=MIGRAPHX` option in CMakeLists.txt (~60 lines of build logic).
- Links against `libmigraphx` (and optional `libmigraphx_gpu`) from rocm.
- Registered as `"mgx"` backend prefix in `setup.cpp`, forced NCHW format.
- Backend identification in `main.cpp`: prints `"Using MIGraphX backend"`.

### Known Limitations
- **No dynamic batch**: Fixed batch sizes require pre-compilation; small batches may waste GPU cycles (e.g., actual batch 3 uses compiled batch 4).
- **NCHW only**: NHWC format is silently ignored.
- **SGF Metadata Encoder disabled**: Potential weight shape mismatch, skipped for now.
- **No mask support in global pooling**: Assumes full board at `nnXLen×nnYLen`.

## Change Stats — ROCm
- Commits: `23` (`non-merge: 18`, `merge: 5`) + post-PR bug fixes
- Files touched: `21` + `3` (rocmbackend.cpp, `rocmhelpers.hip`, `rocmhelpers.h`)
- Diffstat: `+9372 / -4009` + `+59 / -28`

## Change Stats — MIGraphX
- Commits: `3` (`non-merge: 3`)
- Files touched: `4` (migraphxbackend.cpp, CMakeLists.txt, setup.cpp, main.cpp)
- Diffstat: `+1977 / -1` (1886 lines new backend + 91 lines build/setup integration)

### ROCm Backend (`AMD_GPU` branch, 2025-07-28 ~ 2026-03-16)
- `1f2ae46e` 2025-07-28 Add ROCm backend
- `b4555304` 2025-07-28 Fix bugs
- `8b30cb96` 2025-07-31 Update
- `570ced01` 2025-08-01 Fix bugs
- `abb61240` 2025-08-01 Fix bugs
- `bfb292e7` 2025-08-01 All bug fixed
- `4606424f` 2025-08-01 Update
- `1e8ea788` 2025-08-02 test new method
- `c1a09cf3` 2025-08-02 Update
- `0957b88b` 2025-08-02 Test finished
- `c70d841a` 2025-08-02 Update docks
- `1d05ca8d` 2025-08-02 Update gitignore
- `9d4662b7` 2025-08-02 Update new method
- `d40bd509` 2025-08-02 Optimize performance
- `158d24df` 2025-08-13 Update new Convlayer method
- `ec32eb19` 2025-08-13 Merge branch 'master' of https://github.com/Looong01/KataGo-ROCm
- `0bfe0a14` 2025-10-04 Add new compile target
- `f5fbb336` 2025-11-08 Merge branch 'lightvector:master' into master
- `26d8c5bd` 2025-11-08 Add ROCm for Windows support
- `555d2f17` 2025-12-01 Merge branch 'lightvector:master' into master
- `dbc7cfa4` 2026-02-22 Merge branch 'lightvector:master' into master
- `ed396b72` 2026-02-28 Fix bugs
- `ccec62c5` 2026-03-16 Merge branch 'lightvector:master' into master
- `xxxxxxxx` 2026-04-19 Fix critical ConvLayer accumulate bug & algorithm buffer overflow

### MIGraphX Backend (`MIGraphX` branch, 2026-02-27 ~ 2026-04-19)
- `c511c338` 2026-02-27 Add MIGraphX support
- `00cb6881` 2026-04-19 Fix bugs (MIGraphX: 5 structural bugs, GELU→MishScale8, NHWC→NCHW, dimension mismatches)
- `b1da0e06` 2026-04-19 Optimize performance (FP16 default, multi-batch compilation, cache per batch size)


Looong01:
**Windows ROCm:**

image


**Linux ROCm:**

rocm


**Linux MIGraphx:**

image

Looong01:
image

lightvector:
Thanks, I'll look at this soon.

Looong01:
> Thanks, I'll look at this soon.

Thanks!

cpp/command/benchmark.cpp
```diff
@@ -265,6 +265,11 @@ int MainCmds::benchmark(const vector<string>& args) {
   cout << "If you have a strong GPU capable of FP16 tensor cores (e.g. RTX2080), "
        << "using the Cuda version of KataGo instead may give a mild performance boost." << endl;
 #endif
+#ifdef USE_ROCM_BACKEND
+  cout << "You are currently using the ROCm version of KataGo." << endl;
+  cout << "If you have a strong GPU capable of FP16 tensor cores (e.g. RX6900XT), "
+       << "using the ROCm version of KataGo instead may give a mild performance boost." << endl;
```
lightvector: Is this a typo? If someone is using the ROCm version already, why do we advise them to use the ROCm vesion again?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    int numValueChannels;
+    int numScoreValueChannels;
+    int numOwnershipChannels;
+    
+    MIGraphXModel()
+        : modelVersion(0), maxBatchSize(1), nnXLen(19), nnYLen(19),
+          useFP16(false), useNHWC(false),
+          numInputChannels(0), numInputGlobalChannels(0), numInputMetaChannels(0),
+          numPolicyChannels(0), numValueChannels(3),
+          numScoreValueChannels(0), numOwnershipChannels(0) {}
```
lightvector: Style: perhaps these should all be required arguments, without having defaults so that it's impossible to typo and forget a to set a value properly. I notice that in the site where we create the model, we immediately assign all these values and overwrite them anyways.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    // Add a convolution layer
+    migraphx::instruction_ref addConv(
+        migraphx::instruction_ref input,
+        const ConvLayerDesc& convDesc
+    ) {
+        // Validate dimensions
+        if(convDesc.inChannels <= 0 || convDesc.inChannels > 10000 ||
+           convDesc.outChannels <= 0 || convDesc.outChannels > 10000 ||
+           convDesc.convYSize <= 0 || convDesc.convYSize > 100 ||
+           convDesc.convXSize <= 0 || convDesc.convXSize > 100) {
```
lightvector: If we're going to check this here, this bound seems a little wide - allowing 100x100 convs. Do we really support this, and can we tighten it to only what's actually needed that we realistically support efficiently?

Also, are there any requirements for the convolution size to be odd, or do we also support even? In the case of even, "padding=SAME" style convolutions aren't well defined, right? See padX and padY calculation below that semantically don't make sense in case of even.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+            cerr << "ERROR: Conv " << convDesc.name << " has invalid dimensions (in=" << convDesc.inChannels
+                 << ", out=" << convDesc.outChannels << ", ky=" << convDesc.convYSize 
+                 << ", kx=" << convDesc.convXSize << ")" << endl;
+            return input;
```
lightvector: Why merely print to cerr and return input? This seems like it should be a hard failure rather than just proceeding. Throw StringError with this message?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+        size_t expectedWeights = (size_t)convDesc.outChannels * (size_t)convDesc.inChannels 
+                                 * (size_t)convDesc.convYSize * (size_t)convDesc.convXSize;
+        
+        if(convDesc.weights.size() != expectedWeights) {
+            cerr << "ERROR: Conv " << convDesc.name << " weights size mismatch: "
+                 << convDesc.weights.size() << " vs expected " << expectedWeights
+                 << " (out=" << convDesc.outChannels << ", in=" << convDesc.inChannels
+                 << ", ky=" << convDesc.convYSize << ", kx=" << convDesc.convXSize << ")" << endl;
+            return input;  // Return input to avoid crash
+        }
```
lightvector: Why return input to avoid crash? This seems like exactly the case where crashing is preferable - if the size of the weights mismatches, something is seriously wrong and the code should NOT proceed. Throw StringError with this message instead.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    migraphx::instruction_ref addBatchNorm(
+        migraphx::instruction_ref input,
+        const BatchNormLayerDesc& bnDesc
+    ) {
+        // Skip if BN has no channels or invalid weights
+        if(bnDesc.numChannels <= 0 || bnDesc.numChannels > 10000) {
+            cerr << "WARNING: BatchNorm " << bnDesc.name << " has invalid numChannels=" << bnDesc.numChannels 
+                 << ", skipping BN" << endl;
+            return input;
+        }
```
lightvector: Same comment here.  I don't think we want to skip if the number of channels is invalid. Omitting a layer means the computation becomes nonsense.

I'll stop leaving this repeated comment on future instances, but this applies every subsequent site like this across both backends - can you make it so that upon any serious error, validation failure, etc, we throw StringError with the message rather than printing it to cerr and then simply returning?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    
+    // Add MatMul layer
+    migraphx::instruction_ref addMatMul(
+        migraphx::instruction_ref input,
+        const MatMulLayerDesc& matmulDesc,
+        const MatBiasLayerDesc* biasDesc = nullptr
+    ) {
+        // Validate channel counts
+        if(matmulDesc.inChannels <= 0 || matmulDesc.inChannels > 10000 || 
+           matmulDesc.outChannels <= 0 || matmulDesc.outChannels > 10000) {
```
lightvector: I notice a lot of places have validation against channel count > 10000? Is this a value of significance to migraphx or is this a "just in case" sanity check? Can we factor it out as a common top-level constant like "constexpr int MAX_CHANNELS_SANITY_CHECK = 10000"?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+        }
+        // Fallback to relu
+        return main_module->add_instruction(migraphx::make_op("relu"), input);
+    }
+    
+    // Mish activation: x * tanh(softplus(x)) = x * tanh(log(1 + exp(x)))
+    migraphx::instruction_ref addMish(migraphx::instruction_ref input) {
+        auto inputLens = input->get_shape().lens();
+        // softplus(x) = log(1 + exp(x))
+        auto exp_x = main_module->add_instruction(migraphx::make_op("exp"), input);
```
lightvector: The exp is vulnerable to intermediate overflow if the input is too large. Does migraph offer a safe built-in "softplus" activation function that guards against overflow? If so, you should definitely use the built-in softplus. Otherwise, you need to do a conditional that replaces the part that would overflow with the correct asymptotic linear behavior, see https://github.com/lightvector/KataGo/blob/a1d832c693dbed9026bf56552645f824f1c6ece5/cpp/neuralnet/metallayers.swift#L64-L75 or https://github.com/lightvector/KataGo/blob/a1d832c693dbed9026bf56552645f824f1c6ece5/cpp/neuralnet/openclkernels.cpp#L461-L462

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+        auto scaled = main_module->add_instruction(migraphx::make_op("mul"), input, eight);
+        auto thirty = broadcastScalar(30.0f, inputLens);
+        scaled = main_module->add_instruction(migraphx::make_op("min"), scaled, thirty);
+        // softplus(scaled) = log(1 + exp(scaled))
+        auto exp_s = main_module->add_instruction(migraphx::make_op("exp"), scaled);
+        auto ones = broadcastScalar(1.0f, inputLens);
+        auto one_plus_exp = main_module->add_instruction(migraphx::make_op("add"), exp_s, ones);
+        auto softplus = main_module->add_instruction(migraphx::make_op("log"), one_plus_exp);
+        auto tanh_sp = main_module->add_instruction(migraphx::make_op("tanh"), softplus);
+        return main_module->add_instruction(migraphx::make_op("mul"), input, tanh_sp);
```
lightvector: I notice here does the thresholding guard, but will this code run in FP16? If it runs in FP16, then the max 30 is not safe because in FP16 that will already overflow. Prefer if there is a built-in softplus, otherwise this needs to use a variable guard depending on FP16 vs FP32 I think.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    migraphx::instruction_ref addConcat(const vector<migraphx::instruction_ref>& inputs, int64_t axis) {
+        auto concat_op = migraphx::make_op("concat", {{"axis", axis}});
+        return main_module->add_instruction(concat_op, inputs);
+    }
+    
+    // Global pooling producing 3 features per channel.
+    // For trunk/policy: [mean, mean*scale1, max]
+    // For value head:   [mean, mean*scale1, mean*scale2]
+    // Input: [batch, C, H, W], Output: [batch, C*3]
+    // Note: assumes full board (no mask), correct for standard play at nnXLen x nnYLen.
```
lightvector: Good comment. Will this silently degrade if there is nontrivial mask or will it fail loudly? If it will silently degrade and produce bad results, that is unfortunately not acceptable, and it needs to be updated to take into account the mask.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    const ResidualBlockDesc& blockDesc
+) {
+    auto residual = input;
+    
+    // preBN + preActivation
+    auto x = builder.addBatchNorm(input, blockDesc.preBN);
+    x = builder.addActivation(x, blockDesc.preActivation.activation);
+    
+    // regularConv
+    x = builder.addConv(x, blockDesc.regularConv);
```
lightvector: Minor style note - "norm+act+conv" is a very common unit and it may be a slight nice factoring if you introduce a NormActConv helper and use it everywhere appropriate.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    vector<size_t> inputShape = {(size_t)maxBatchSize, (size_t)numSpatialFeatures, (size_t)nnYLen, (size_t)nnXLen};
+    vector<size_t> inputGlobalShape = {(size_t)maxBatchSize, (size_t)numGlobalFeatures};
+    
+    // Input parameters are always float_type (host buffers are float).
+    // If using FP16, we convert to half inside the graph so MIGraphX handles conversion on GPU.
+    auto inputSpatial = main_module->add_parameter("input_spatial", migraphx::shape(migraphx::shape::float_type, inputShape));
+    auto inputGlobal = main_module->add_parameter("input_global", migraphx::shape(migraphx::shape::float_type, inputGlobalShape));
+    
+    // MIGraphX backend uses NCHW format only
+    (void)useNHWC;  // Silently ignore NHWC setting
```
lightvector: Can you fail loudly instead? See https://github.com/lightvector/KataGo/blob/a1d832c693dbed9026bf56552645f824f1c6ece5/cpp/neuralnet/openclbackend.cpp#L556-L559

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+        auto globalBroadcast = main_module->add_instruction(
+            migraphx::make_op("multibroadcast", {{"out_lens", trunkShape}}), globalUnsqueezed);
+        trunk = main_module->add_instruction(migraphx::make_op("add"), trunk, globalBroadcast);
+    }
+    
+    // SGF Metadata encoder (if enabled) - disabled for now due to potential weight shape issues
+    if(trunkDesc.metaEncoderVersion > 0 && numMetaFeatures > 0) {
+        // Skip SGF metadata encoder for now
+        cout << "MIGraphX: SGF Metadata encoder disabled" << endl;
+    }
```
lightvector: This should not be an if statement. This case should raise StringError (not testAssert, and not cout) so that we get a helpful error and a hard failure if we try to use a model that requires this encoder.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    
+    // trunkTipBN + trunkTipActivation
+    trunk = builder.addBatchNorm(trunk, trunkDesc.trunkTipBN);
+    trunk = builder.addActivation(trunk, trunkDesc.trunkTipActivation.activation);
+    
+    // ======== Policy Head ========
+    const PolicyHeadDesc& policyDesc = modelDesc.policyHead;
+    
+    migraphx::instruction_ref policy = trunk;
+    migraphx::instruction_ref policyPass = trunk; // will be overwritten
```
lightvector: Minor: Does migraph not allow declaring an instruction_ref without assigning it? Why assign and then comment that it's overwritten?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    
+    // Build trunk
+    auto trunk = inputSpatial;
+    const TrunkDesc& trunkDesc = modelDesc.trunk;
+    
+    // Initial conv
+    if(trunkDesc.initialConv.outChannels > 0 && trunkDesc.initialConv.inChannels == numSpatialFeatures) {
+        trunk = builder.addConv(trunk, trunkDesc.initialConv);
+    } else if(trunkDesc.initialConv.outChannels > 0) {
+        cout << "MIGraphX: Skipping initialConv (input channel mismatch)" << endl;
```
lightvector: Style note: we can't print to cout here because some protocols (GTP, analysis engine) mandate that stdout follow a specific protocol. Even cerr won't show up in the log files katago produces, so if something is genuinely a non-fatal warning,  then to log properly you thread through the `Logger* logger` from `createComputeContext`. However, cases like this should just be an error and simply throw StringError.

Here and everywhere else we print to cout or cerr - can you avoid doing so and either raise, or use the Logger for proper logging?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    }
+    
+    // Build trunk
+    auto trunk = inputSpatial;
+    const TrunkDesc& trunkDesc = modelDesc.trunk;
+    
+    // Initial conv
+    if(trunkDesc.initialConv.outChannels > 0 && trunkDesc.initialConv.inChannels == numSpatialFeatures) {
+        trunk = builder.addConv(trunk, trunkDesc.initialConv);
+    } else if(trunkDesc.initialConv.outChannels > 0) {
```
lightvector: This should not be an if statement - if the initial conv channels is an unexpected number, it should be a hard failure (testAssert, or raise StringError) rather than simply skipping the operation.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    if(trunkDesc.initialMatMul.outChannels > 0) {
+        auto globalProcessed = builder.addMatMul(inputGlobal, trunkDesc.initialMatMul);
+        // Broadcast global features from [N, C] to spatial dimensions [N, C, H, W]
+        auto trunkShape = trunk->get_shape().lens();
+        auto globalUnsqueezed = main_module->add_instruction(
+            migraphx::make_op("unsqueeze", {{"axes", migraphx::value(vector<int64_t>{2, 3})}}), globalProcessed);
+        auto globalBroadcast = main_module->add_instruction(
+            migraphx::make_op("multibroadcast", {{"out_lens", trunkShape}}), globalUnsqueezed);
+        trunk = main_module->add_instruction(migraphx::make_op("add"), trunk, globalBroadcast);
+    }
```
lightvector: This should not be an if statement - if the initial matmul channels is an unexpected number, it should be a hard failure (testAssert, or raise StringError) rather than simply skipping the operation.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    trunk = builder.addBatchNorm(trunk, trunkDesc.trunkTipBN);
+    trunk = builder.addActivation(trunk, trunkDesc.trunkTipActivation.activation);
+    
+    // ======== Policy Head ========
+    const PolicyHeadDesc& policyDesc = modelDesc.policyHead;
+    
+    migraphx::instruction_ref policy = trunk;
+    migraphx::instruction_ref policyPass = trunk; // will be overwritten
+    
+    if(policyDesc.p1Conv.outChannels > 0) {
```
lightvector: This should not be an if statement. This case should either testAssert or raise StringError if the channels is an unexpected number and then the code inside should run unconditionally. This might also simplify the above issue about migraphx::instruction_ref needing an initial assignment.

I'll stop leaving comments like this too but this applies to every situation like this across both backends - if something is not supported, it should NOT simply proceed silently and risk outputting garbage.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    
+    // Score value branch: same v2 -> sv3Mul + sv3Bias
+    auto scoreValue = builder.addMatMul(v2, valueDesc.sv3Mul, &valueDesc.sv3Bias);
+    
+    // Set outputs: policy, policyPass, value, scoreValue, ownership
+    if(modelDesc.modelVersion >= 2) {
+        main_module->add_return({policy, policyPass, valueOut, scoreValue, ownership});
+    } else {
+        main_module->add_return({policy, valueOut});
+    }
```
lightvector: I'm confused about the modelVersion >= 2 check here, this is matched with no other backends I think? modelVersion == 2 isn't even supported any more, so it seems like this is a typo? What version check was intended here?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+        delete context;
+    }
+}
+
+// Static mutex for cache operations
+static mutex migraphxCacheMutex;
+
+// Generate batch sizes to compile for MIGraphX (no dynamic batch support).
+static vector<int> generateBatchSizes(int maxBatchSize) {
+    vector<int> candidates = {4, 8, 16, 24, 32, 40, 64};
```
lightvector: No dynamic batch support is interesting! What's the practical cost of having more entries? Is it more GPU memory for storing copies of the weight tensors or things like that, or is that all shared and it's just a bit of extra memory on the CPU side for the graph structures, or something else?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+        "migraphx_%s_%s_%dx%d_batch%d_fp%d_nhwc%d_%s",
+        modelDesc.name.c_str(),
+        modelDesc.sha256.substr(0, 16).c_str(),
+        nnYLen,
+        nnXLen,
+        maxBatchSize,
+        useFP16 ? 1 : 0,
+        useNHWC ? 1 : 0,
+        requireExactNNLen ? "exact" : "max"
+    );
```
lightvector: Does the cache contain the entire copy of the model weights? (This could build up to a large amount of disk space over time if people keep running new models, such as for rating games during contribute, so if that's the case it's something to warn about in documentation perhaps).

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    handle->nnYLen = ctx->nnYLen;
+    
+    bool useFP16 = (ctx->useFP16Mode == enabled_t::True || ctx->useFP16Mode == enabled_t::Auto);
+    bool useNHWC = (ctx->useNHWCMode == enabled_t::True);
+    
+    // MIGraphX backend only supports NCHW format
+    if(useNHWC) {
+        cout << "MIGraphX: WARNING: NHWC format is not supported, forcing NCHW" << endl;
+        useNHWC = false;
+    }
```
lightvector: See prior comment about NHWC. Prefer to do what OpenCL does and fail loud, and also avoid printing to cout as explained in a prior comment.

lightvector: Same for all the printing to cout below in the function and in the rest of this backend or the other backend.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+            handle->model->progs[bs].compile(handle->model->tgt, compile_opts);
+            
+            cout << "MIGraphX: Batch " << bs << " compiled!" << endl;
+            
+            // Save to cache
+            try {
+                if(logger) {
+                    logger->write("MIGraphX: Saving compiled program to cache (batch " + Global::intToString(bs) + "): " + cacheFile);
+                }
+                migraphx::save(handle->model->progs[bs], cacheFile);
```
lightvector: Question: What happens if two separate instances of KataGo are launched concurrently by a user for the same model and they both try to save over the same file here concurrently?

(I don't think other backends have any particular good handling of this kind of issue either, but it occured to me while examining the code here and it may be worth considering)

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+            output->shorttermWinlossError = 0.0f;
+            output->shorttermScoreError = 0.0f;
+        } else {
+            output->whiteScoreMean = 0.0f;
+            output->whiteScoreMeanSq = 1.0f;
+            output->whiteLead = 0.0f;
+            output->varTimeLeft = 0.0f;
+            output->shorttermWinlossError = 0.0f;
+            output->shorttermScoreError = 0.0f;
+        }
```
lightvector: Most other backends have this case as ASSERT_UNREACHABLE.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+        results[0].visit([&](auto output) {
+            for(int row = 0; row < batchSize; row++) {
+                for(size_t i = 0; i < policySize; i++) {
+                    buffers->policyResults[row * policySize + i] = static_cast<float>(output[row * policySize + i]);
+                }
+            }
+        });
+    }
+    
+    if(modelVersion >= 2) {
```
lightvector: Other backends don't have a check for modelVersion >= 2 here. Why do we need it here?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+) {
+    auto cacheDir = HomeData::getHomeDataDir(true, homeDataDir);
+    cacheDir += "/migraphxcache";
+    
+    // Create directory if not exists
+    MakeDir::make(cacheDir);
+    
+    // Generate unique cache key based on model and parameters
+    string cacheKey = Global::strprintf(
+        "migraphx_%s_%s_%dx%d_batch%d_fp%d_nhwc%d_%s",
```
lightvector: Does the cache remain legitimate if migraph itself changes versions? And even if it remains legitimate, if migraph releases improvements to its graph compiler in a later version, will the caching mean users fail to pick up those improvements?

I think this key needs to include more components, to ensure it invalidates properly and re-computes in such cases.

cpp/main.cpp
```diff
@@ -278,6 +287,8 @@ string Version::getGitRevisionWithBackend() {
   s += "-cuda";
 #elif defined(USE_TENSORRT_BACKEND)
   s += "-trt";
+#elif defined(USE_ROCM_BACKEND)
+  s += "-rocm";
```
lightvector, line 294: Migraph is missing here?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+                bs,
+                ctx->nnXLen,
+                ctx->nnYLen,
+                useFP16,
+                useNHWC
+            );
+            
+            cout << "MIGraphX: Compiling batch " << bs << "..." << endl;
+            migraphx::compile_options compile_opts;
+            compile_opts.offload_copy = true;
```
lightvector: Can you explain what offload_copy does? I'm not so familiar with migraph

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    int serverThreadIdx
+) {
+    (void)serverThreadIdx;
+    
+    ComputeContextInternal* ctx = reinterpret_cast<ComputeContextInternal*>(context);
+    const LoadedModelInternal* model = reinterpret_cast<const LoadedModelInternal*>(loadedModel);
+    
+    auto handle = new ComputeHandleInternal();
+    handle->maxBatchSize = maxBatchSize;
+    handle->gpuIdx = gpuIdxForThisThread;
```
lightvector: Is there somewhere in this backend where this is ever used? Does migraph simply ignore the gpu idx directives from the user and use whatever GPUs it wants? That seems not ideal if the case, but I might just be missing where you use it here in this code?

cpp/neuralnet/migraphxbackend.cpp
lightvector: Oh, neat, apparently the following identity exists:
```
softplus(x) = max(x,0) + log1p(exp(-|x|))
```
I might see if it would simpilify other backends to use this identity as well.

cpp/command/benchmark.cpp
```diff
@@ -265,6 +265,11 @@ int MainCmds::benchmark(const vector<string>& args) {
   cout << "If you have a strong GPU capable of FP16 tensor cores (e.g. RTX2080), "
        << "using the Cuda version of KataGo instead may give a mild performance boost." << endl;
 #endif
+#ifdef USE_ROCM_BACKEND
```
lightvector, line 267: Is migraph missing from here?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    // Input: [batch, C, H, W], Output: [batch, C*3]
+    // Note: assumes full board (no mask), correct for standard play at nnXLen x nnYLen.
+    migraphx::instruction_ref addGPool(migraphx::instruction_ref input, bool isValueHead = false) {
+        float boardArea = (float)(nnXLen * nnYLen);
+        float sqrtBoardArea = sqrtf(boardArea);
+        float scale1Factor = (sqrtBoardArea - 14.0f) * 0.1f;
+        
+        // mean: [batch, C, H, W] -> [batch, C, 1, 1] -> [batch, C]
+        auto mean = addReduceMean(input, {2, 3});
+        mean = addSqueeze(mean, {2, 3});
```
lightvector: Does this need a keepdims or something?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+// - BatchNorm is simplified (skipped) due to MIGraphX broadcast limitations
+// - Global pooling residual blocks use simplified implementation
+// - Value/Score/Ownership heads use simplified projections
+//
+// Future Optimizations:
+// - Implement proper BatchNorm with broadcast
+// - Full global pooling residual block implementation
+// - Complete value head with v2Mul/v3Mul layers
+// - FP16 support for faster inference
+//
```
lightvector: Is this doc up to date? For example, it claims FP16 is future optimizations - do we not support FP16 here?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
...
+    // Add residual
+    return builder.main_module->add_instruction(migraphx::make_op("add"), x, residual);
+}
+
+// Forward declarations
+static migraphx::instruction_ref buildResidualBlock(
+    MIGraphXGraphBuilder& builder,
+    migraphx::instruction_ref input,
+    const ResidualBlockDesc& blockDesc
+);
```
lightvector: This forward declaration occurs after its real declaration - is it needed?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1886 @@
+#include "../neuralnet/nninterface.h"
+#include "../neuralnet/nninputs.h"
+#include "../neuralnet/nneval.h"
+#include "../neuralnet/modelversion.h"
+#include "../neuralnet/desc.h"
+#include "../neuralnet/sgfmetadata.h"
+#include "../neuralnet/activations.h"
+#include "../neuralnet/activations.h"
```
lightvector: Included twice?

lightvector (changes requested):
Thanks for all this work and sorry for the delay in reading over this! I've left a bunch of comments about migraph. I haven't done a detailed review of ROCm yet, but I suspect a lot of the high-level comments about error handling I left about migraph will apply to ROCm yet, can you take a look?

cpp/neuralnet/migraphxbackend.cpp
Looong01: Fixed.

cpp/neuralnet/migraphxbackend.cpp
Looong01: Fixed.

Looong01:
Thank you so much for seeing this. I have fixed all and answered part of your questions about MIGraphX backend.
And There is only one similar issue in ROCm backend(`cout << "Found ROCm device " << i << ": " << prop.name << endl;`). Now I fix it.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1828 @@
...
+            );
+        auto weights = addLiteral(matmulDesc.weights, wShape);
+        
+        auto matmul = main_module->add_instruction(migraphx::make_op("dot"), input, weights);
+        
+        if(biasDesc != nullptr && !biasDesc->weights.empty()) {
+            if(biasDesc->weights.size() != (size_t)biasDesc->numChannels) {
+                cerr << "ERROR: MatMul bias " << biasDesc->name << " size mismatch: "
+                     << biasDesc->weights.size() << " vs expected " << biasDesc->numChannels << endl;
+            } else {
```
lightvector: Missed spot in earlier error handling cleanup.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1828 @@
...
+    // Initial conv
+    if(trunkDesc.initialConv.inChannels != numSpatialFeatures)
+        throw StringError(
+            "MIGraphX: initialConv input channels mismatch: expected " + Global::intToString(numSpatialFeatures) +
+            " but got " + Global::intToString(trunkDesc.initialConv.inChannels)
+        );
+    trunk = builder.addConv(trunk, trunkDesc.initialConv);
+
+    // Initial MatMul for global features
+    {
```
lightvector: It might be worth an error check that the initial matmul channels matches the numGlobalFeatures.

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1828 @@
...
+    
+    // Create directory if not exists
+    MakeDir::make(cacheDir);
+    
+    // Cache key includes MIGraphX version to invalidate when the compiler changes
+#if defined(MIGRAPHX_VERSION_MAJOR) && defined(MIGRAPHX_VERSION_MINOR) && defined(MIGRAPHX_VERSION_PATCH)
+    string migraphxVersionStr = Global::strprintf("%d_%d_%d", MIGRAPHX_VERSION_MAJOR, MIGRAPHX_VERSION_MINOR, MIGRAPHX_VERSION_PATCH);
+#else
+    string migraphxVersionStr = "unknown";
+#endif
```
lightvector: If they aren't defined, what are your thoughts about emitting a warning through the logger and not saving the cache file?

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1828 @@
...
+    string cacheKey = Global::strprintf(
+        "migraphx%s_%s_%s_%dx%d_batch%d_fp%d_%s",
```
lightvector: Should the cache also include something based on GPU architecture/name? ("gcnArchName"?)

cpp/neuralnet/migraphxbackend.cpp
```diff
@@ -0,0 +1,1828 @@
...
+        
+        // preBN + preActivation (simplified - just activation for now)
+        auto x = input;
+        if(desc->preActivation.activation == 1) {  // GELU
+            // Simplified GELU
+            auto sigmoid = main_module->add_instruction(migraphx::make_op("sigmoid"), x);
+            x = main_module->add_instruction(migraphx::make_op("mul"), x, sigmoid);
+        } else {
+            x = main_module->add_instruction(migraphx::make_op("relu"), x);
+        }
```
lightvector: To what degree is it possible to make these these tests exercise the actual same classes or components of the graph logic for the actual residual blocks, conv layers, etc, by calling the same classes or functions to build them?

For example I notice here there is custom "gelu" code that doesn't appear anywhere else and definitely is not part of the actual residual block implementation.

cpp/CMakeLists.txt
```diff
@@ -496,6 +869,35 @@ elseif(USE_BACKEND STREQUAL "EIGEN")
...
+    target_link_libraries(katago amdhip64)
+  endif()
+  
+  # Link other required libraries
+  find_library(HIPRTC_LIBRARY hiprtc
+               HINTS /opt/rocm/lib
+               PATH_SUFFIXES lib lib64)
+  if(HIPRTC_LIBRARY)
+    target_link_libraries(katago ${HIPRTC_LIBRARY})
+  endif()
```
lightvector: If some of these libraries are required, would it be better to report an error if they're not found and/or fail here with a message that flags why, rather than waiting for a link-time error?

cpp/CMakeLists.txt
```diff
@@ -496,6 +869,35 @@ elseif(USE_BACKEND STREQUAL "EIGEN")
...
+  
+  # Add ROCm library directories
+  link_directories(/opt/rocm/lib)
```
lightvector: Is this a no-op given line 428, or am I misreading the logic?

lightvector (commented):
Thanks for the fixes. I marked as resolved the comments that looked resolved to me but left open the ones that I didn't see an answer to, or maybe I just missed it. Please let me know what you think about them - you can reply back to the comments if you have an an answer or you think a resolution is not necessary and I will take a look.

I also left some more comments in this pass.

I have a much higher-level question too - the MIGraphX backend looks quite reasonable and I think I'm okay accepting it once it's polished and once I can independently test it a little too. 

However, the ROCm backend I'm much less sure about. I skimmed through it, and it seems like a massive copy-paste of the CUDA backend, but with lots of subtle differences, and that seems really awkward for long-term maintainability. Do you have thoughts about this? How necessary is this backend if we have both OpenCL and MIGraphX - do you have thoughts on if it would be okay to drop it, or split it to a separate PR, or keep it as an unofficial branch or something, or other options?

Additionally, very soon I'm going to be adding support for transformer blocks, since transformers are likely the future of strong models. I have a branch where I'm working on it for OpenCL and testing it now, and will be implementing personally it through Eigen/CUDA/TensorRT as well. What are your thoughts about implementing support for these for the AMD side here once it's ready as well? Is this something you're committed to maintaining or updating through more architecture changes?

Looong01:
In fact, ROCm is more mature than MIGraphX. This can be seen through benchmarks: the computing speed of ROCm is greater than MIGraphX and far greater than OpenCL.

ROCm itself is designed to be compatible with CUDA. I don't think this is very awkward for long-term maintenance, because code can be very easily and effortlessly migrated from CUDA to ROCm. Moreover, ROCm supports both Linux and Windows, while MIGraphX currently officially only supports Linux. According to AMD's official roadmap, MIGraphX may support Windows much later, although Windows support is part of their plan.

In addition, ROCm and MIGraphX support all CUDA features and operators, including but not limited to Transformer blocks. If you are ready to add any new features, I will update in a timely manner to add the corresponding AMD GPU support for Transformers or any new features in the future.

lightvector:
Huh, I didn't know that ROCm was a lot faster. What about I focus on reviewing that, and we simply drop MIGraphX support entirely then? I don't want to keep introducing tons of backends especially if it's less compatible and slower right now.

Looong01:
> Huh, I didn't know that ROCm was a lot faster. What about I focus on reviewing that, and we simply drop MIGraphX support entirely then? I don't want to keep introducing tons of backends especially if it's less compatible and slower right now.

Well, i make a mistake that ROCm is faster than MIGraphX but not a lot faster. Because some ops of MIGraphX do not support dynamic shape. So I had to compile multiple static .mxr model to do compute. And this make it slower.

I agree with u. So I detach MIGraphX backend from ROCm now. I will only PR ROCm and MIGraphX will become a standlone branch until it support dynamic shape(Full dynamic shape support is still in the design discussion stage (#2046) and is not a delivered feature.).

Looong01:
# Add ROCm support for transformer/attention models, optional CK fused attention, and multi-arch build robustness fixes（Just like fused attetion of CUDA backend）

## Summary

The upstream `4636f6bf` merge added transformer/attention blocks (model version 17: RMSNorm,
RoPE, GQA, SwiGLU) to the CUDA backend but never touched ROCm, leaving it unable to compile
(shared `NeuralNet` interface changed) and unable to load v17 models at all. This PR brings the
ROCm backend up to parity and adds an optional fused-attention fast path, plus fixes several
build-robustness issues discovered along the way.

- **Interface fix**: update `rocmbackend.cpp` to the new `createComputeContext`/`setIsWarmup`
  interface (this alone was blocking all ROCm builds).
- **Transformer/attention support**: port RMSNorm (non-spatial + spatial), RoPE (fixed +
  learnable), SwiGLU, and a plain online-softmax attention kernel to `rocmhelpers.hip`; wire up
  `TransformerAttentionBlock`/`TransformerFFNBlock`/`RMSNormLayer` in `rocmbackend.cpp`, mirroring
  the CUDA backend's implementation. ROCm already had full NHWC plumbing that was previously
  disabled at the entry point; transformer models now auto-enable it (matching CUDA's behavior).
- **Optional CK fused-attention path**: vendors a small set of glue headers + pre-generated kernel
  instantiations from AMD's Composable Kernel (`ck_tile` FMHA) under
  `cpp/external/composable_kernel_fmha/`, mirroring the CUDA backend's optional cudnn-frontend
  SDPA path. Falls back gracefully to the plain kernel if a compatible CK version isn't installed,
  if the shape/arch isn't supported (RDNA1/RDNA2 have no CK FMHA kernels), or if disabled via
  `rocmDisableFusedAttention = true`. ***Measured ~2.2x nnEvals/s on a gfx1100 (RX 7900 XTX) on Linux, 
  ~1.3x faster on Windows on the same GPU (Windows' WDDM driver model adds overhead CUDA/ROCm
  don't see on Linux).***
- **Build robustness fixes** (found while getting the above working):
  - Resolve the real ROCm install prefix (`/opt/rocm/core-<version>/`, used since ROCm ~7.9,
    highest version preferred; falls back to the old flat `/opt/rocm/` layout) instead of assuming
    `/opt/rocm` directly - drives compiler detection, library/include search, and now the binary's
    baked-in RPATH (previously unset, so the built binary depended on `/opt/rocm` continuing to
    point at the same install and/or `LD_LIBRARY_PATH` being set at run time).
  - Fix `CMAKE_HIP_COMPILER` detection to always use `clang++` directly (CMake rejects the `hipcc`
    wrapper script) and probe for both `hipcc`/`clang++` independently instead of one gating the
    other.
  - Fix a real bug in the pre-project multi-arch default logic: `project(... HIP)` triggers
    CMake's own HIP detection which populated `CMAKE_HIP_ARCHITECTURES` with just the native GPU
    before our broad-default logic ran, making it a no-op on Linux.
  - Fix `-I` vs `-isystem` include ordering for HIP headers: a newer compiler paired with a stale
    system `libamdhip64-dev`'s headers under `/usr/include` broke compilation; plain `-I` lost to
    `/usr/include`, `-isystem` doesn't.
  - Expand `katago_default_hip_archs()` from 10 to 25 real AMD GPU architectures (was missing all
    of RDNA1, most of RDNA2/RDNA3/RDNA3.5, and RDNA4 entirely).

## Test plan
- [x] Built successfully with `-DUSE_BACKEND=ROCM` (both single-arch and the full 25-arch fat
      binary), no explicit `-D` flags required for a fresh checkout.
- [x] Regression-tested an existing pre-v17 model (`g170-b6c96-...`) - no change in behavior.
- [x] Ran `testgpuerror` (fp16 vs fp32 cross-check) against all three new v17 test models
      (RoPE+RMSNorm, GQA+learnable-RoPE+BatchNorm-tip, NBT+SwiGLU+spatial-RMSNorm) on a real
      gfx1100 GPU, both with and without the CK fused path, and on rectangular boards - FP16 error
      stayed within normal precision-loss range in all cases.
- [x] Verified the built binary's RPATH and confirmed it runs with no `LD_LIBRARY_PATH` override.
- [x] Benchmarked CK fused attention vs the plain kernel (~2.2x nnEvals/s improvement).

@lightvector

lightvector:
@Looong01 Thanks for all the patience here. This PR has now been squashed and merged as dc7dd784 into https://github.com/lightvector/KataGo/pull/1234 which will go right into master if it seems to be working well. Starting with all your changes as a solid foundation to test from, I further made some refactors to fix some remaining issues and then unify this backend with the CUDA backend, see b90b1bb5 and c002c973 and ec7470ea mostly.

I'm very uncertain whether some of the changes to the cmake logic are still all working, although I did try to test myself on a rented linux amd gpu machine, and minimally test the build on a windows machine (but no actual amd gpu).

Would you be able to take a look and to test that the tip of #1234 is working for you and performs the way you expect and so on? If we can confirm that rocm and onnx backends are both working on the tip (onnx backend was also folded into the same PR) then this PR become the new master tip and we should be good to go I hope.

Looong01:
> @Looong01 Thanks for all the patience here. This PR has now been squashed and merged as [dc7dd78](https://github.com/lightvector/KataGo/commit/dc7dd7840373ff6387816bdc46dcd81bb2106680) into #1234 which will go right into master if it seems to be working well. Starting with all your changes as a solid foundation to test from, I further made some refactors to fix some remaining issues and then unify this backend with the CUDA backend, see [b90b1bb](https://github.com/lightvector/KataGo/commit/b90b1bb556a6494afe4f64f2517c314642749501) and [c002c97](https://github.com/lightvector/KataGo/commit/c002c97337ae1a6f39fa65995b0dd1c792f7252f) and [ec7470e](https://github.com/lightvector/KataGo/commit/ec7470eafc7cf85196f3d4c1161cbc1a079bb275) mostly.
> 
> I'm very uncertain whether some of the changes to the cmake logic are still all working, although I did try to test myself on a rented linux amd gpu machine, and minimally test the build on a windows machine (but no actual amd gpu).
> 
> Would you be able to take a look and to test that the tip of #1234 is working for you and performs the way you expect and so on? If we can confirm that rocm and onnx backends are both working on the tip (onnx backend was also folded into the same PR) then this PR become the new master tip and we should be good to go I hope.

Hi, I just do a upgrade for v1.17.2. Just including a very little change: https://github.com/lightvector/KataGo/pull/1188/commits/86c7aaa6d3b8b83f60c2dcfec552bb5ba330137c
Maybe u can have a look!

Looong01:
> @Looong01 Thanks for all the patience here. This PR has now been squashed and merged as [dc7dd78](https://github.com/lightvector/KataGo/commit/dc7dd7840373ff6387816bdc46dcd81bb2106680) into #1234 which will go right into master if it seems to be working well. Starting with all your changes as a solid foundation to test from, I further made some refactors to fix some remaining issues and then unify this backend with the CUDA backend, see [b90b1bb](https://github.com/lightvector/KataGo/commit/b90b1bb556a6494afe4f64f2517c314642749501) and [c002c97](https://github.com/lightvector/KataGo/commit/c002c97337ae1a6f39fa65995b0dd1c792f7252f) and [ec7470e](https://github.com/lightvector/KataGo/commit/ec7470eafc7cf85196f3d4c1161cbc1a079bb275) mostly.
> 
> I'm very uncertain whether some of the changes to the cmake logic are still all working, although I did try to test myself on a rented linux amd gpu machine, and minimally test the build on a windows machine (but no actual amd gpu).
> 
> Would you be able to take a look and to test that the tip of #1234 is working for you and performs the way you expect and so on? If we can confirm that rocm and onnx backends are both working on the tip (onnx backend was also folded into the same PR) then this PR become the new master tip and we should be good to go I hope.

And I am sure everything is OK to merge! My test result & report as following:

Tested the tip of #1234 (66b12ac3) on my machine. ROCm backend builds and runs.

**Environment:** RX 7900 XTX (gfx1100), ROCm 7.14, MIOpen 3.5.2, Composable Kernel 1.2.0, clang 23, Linux x86-64.

**Build:** Clean configure + full build (`-DUSE_BACKEND=ROCM -DCMAKE_BUILD_TYPE=Release`, 25 archs, `make -j8`). The ROCm prefix resolution correctly picks `/opt/rocm/core-7.14`, CK FMHA fused attention is enabled, and RPATH is set correctly. `katago version` reports v1.17.2 + 66b12ac3 with the ROCm backend.

**Functionality:** All four models load and play via GTP (genmove returns sane moves): b10c384h6nbttflrs, b10c512h8nbt3tflrs-fson-silu-rsnh, b11c768h12nbt3tflrs-fson-silu, kata1-zhizi-b40c768nbt-s11272M-d5935M.

**Benchmark** (peak visits/s, back-to-back on the same GPU, vs my AMD_GPU branch at v1.17.2 which is what dc7dd78 was squashed from):

| Model | My branch | #1234 tip | Delta |
|---|---|---|---|
| b10c384h6nbttflrs | 2490 | 2649 | +6.4% |
| b10c512h8nbt3tflrs-fson-silu-rsnh | 1396 | 1506 | +7.9% |
| kata1-zhizi-b40c768nbt-s11272M-d5935M | 336 | 354 | +5.4% |
| b11c768h12nbt3tflrs-fson-silu | 927 | 780 | **-15.8%** |

The first three are faster than before, which is great. However, the largest transformer (b11c768h12, also the flagship model of the three new ones) shows a reproducible ~15% regression (reproduced across two separate runs, so not thermal drift).

Some isolation I did on the b11 regression:
- `rocmInputsUseNHWC=false` gives the same 782 v/s, so the inputsUseNHWC default flip for ROCm in c002c97 is not the cause.
- With `rocmDisableFusedAttention=true`: my branch drops to 576, #1234 drops to 493. So the unified backend is ~15% slower for this model on BOTH the plain attention path and the CK FMHA path (my branch 576 -> 927 with CK on, #1234 493 -> 780 with CK on). The regression is not in the CK dispatch but somewhere in the shared trunk path (conv/matmul/buffer handling or MIOpen solution selection) that only hurts the 768-channel pure transformer trunk — the 384/512-channel transformers and the 768-channel hybrid b40 are all unaffected (and faster).

But even with this performance regression, overall it looks good to me and I think it's worth merging.

lightvector:
I believe this PR as it is now actually merged in to master as https://github.com/lightvector/KataGo/commit/dc7dd7840373ff6387816bdc46dcd81bb2106680, (squashed, as part of https://github.com/lightvector/KataGo/pull/1234) 
with the prior mentioned additional refactors and work I did after that stacked on top.

@Looong01 - thanks for all the testing and coding work. If you agree, you can close this PR.

I would be very interested to know what the 15% slowdown on the largest model is from, if you still have the time and desire to test it out. I don't think on my test cloud machine I noticed a significant slowdown from the changes, and I don't have your hardware to replicate and bisect, but if you have the ability to do so maybe it would reveal some further optimizations that could be very helpful.

If you do have further work or optimizations or figure out what the 15% is from, please feel free to open a new PR against the new master branch.

Looong01:
> I believe this PR as it is now actually merged in to master as [dc7dd78](https://github.com/lightvector/KataGo/commit/dc7dd7840373ff6387816bdc46dcd81bb2106680), (squashed, as part of #1234) with the prior mentioned additional refactors and work I did after that stacked on top.
> 
> @Looong01 - thanks for all the testing and coding work. If you agree, you can close this PR.
> 
> I would be very interested to know what the 15% slowdown on the largest model is from, if you still have the time and desire to test it out. I don't think on my test cloud machine I noticed a significant slowdown from the changes, and I don't have your hardware to replicate and bisect, but if you have the ability to do so maybe it would reveal some further optimizations that could be very helpful.
> 
> If you do have further work or optimizations or figure out what the 15% is from, please feel free to open a new PR against the new master branch.

Thank u so much! I will have a look on this slowdown.

PR #1190, opened 2026-04-25
Configure clang-tidy static analysis in CMake

KvanTTT:
Enable already applied checks (from #1167)

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -50,6 +50,8 @@ set(USE_BIGGER_BOARDS_EXPENSIVE 0 CACHE BOOL "Allow boards up to size 50. Compil
 set(USE_CACHE_TENSORRT_PLAN 0 CACHE BOOL "Use TENSORRT plan cache. May use a lot of disk space. Only applies when USE_BACKEND is TENSORRT.")
 mark_as_advanced(USE_CACHE_TENSORRT_PLAN)
 
+set(USE_CLANG_TIDY 1 CACHE BOOL "Run clang-tidy static analysis during build")
+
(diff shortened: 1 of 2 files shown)


KvanTTT:
For some reason, the warnings are not being reported by MSVC compiler (but it works locally), but they are being reported on macOS and Linux: https://github.com/lightvector/KataGo/actions/runs/24933559965/job/73015201820?pr=1190

I recommend fixing/suppressing remaining and new warnings (for instance, introduced by ed71a1fdd8b6f28a8bb274ebdd690bb484d2a7b2)

cpp/CMakeLists.txt
```diff
@@ -610,5 +612,37 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "C
...
+  find_program(CLANG_TIDY_EXE NAMES clang-tidy HINTS ${CLANG_TIDY_HINTS} REQUIRED)
+
+  if(MSVC)
+    set(CLANG_TIDY_EXTRA_ARGS "--extra-arg=/EHsc")
+  else()
+    set(CLANG_TIDY_EXTRA_ARGS "--extra-arg=-fexceptions")
+  endif()
+
+  set_target_properties(katago PROPERTIES
+    CXX_CLANG_TIDY "${CLANG_TIDY_EXE};-checks=-*,performance-inefficient-vector-operation,performance-unnecessary-copy-initialization,readability-non-const-parameter,readability-make-member-function-const;${CLANG_TIDY_EXTRA_ARGS}")
```
KvanTTT: We probably could enable all performance-related lints:

```
performance-*
```

But it provokes too many warnings and most of them aren't valuable.

PR #1203, opened 2026-05-31
fix: set optimizer_name in train_state when loading checkpoint without optimizer

loker404:
fix: set optimizer_name in train_state when loading checkpoint without optimizer


lightvector:
Thanks!

PR #1204, opened 2026-05-31
CI: key macOS-metal CMake cache on Homebrew dependency versions

ChinChangYang:
## Summary

Split out from #1202 as a standalone, orthogonal CI fix.

The `build-macos-metal` job caches `CMakeCache.txt`/`build.ninja` keyed only on `CMakeLists.txt` hashes. Those cached files bake in version-pinned Homebrew Cellar paths emitted by pkg-config at configure time (e.g. `-L/opt/homebrew/Cellar/protobuf/<ver>/lib`). When Homebrew bumps `protobuf`/`abseil`, those paths vanish, and an incremental `cmake .` does **not** refresh the cached pkg-config vars — so the re-link fails with `ld: library 'protobuf' not found`.

## Fix

- Resolve the installed `protobuf`/`abseil` versions into a `depvers` step output.
- Include those versions in the cache key and scope `restore-keys` to them, so a dependency bump invalidates the stale configure and forces a fresh pkg-config resolution instead of resurrecting dead Cellar paths.

Touches only `.github/workflows/build.yml`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,6 +110,18 @@ jobs:
         run: |
           brew install ninja zlib libzip protobuf abseil
 
+      # The cached CMakeCache.txt/build.ninja bake in version-pinned Homebrew
+      # Cellar paths (e.g. -L/opt/homebrew/Cellar/protobuf/<ver>/lib) that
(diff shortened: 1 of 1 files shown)


ChinChangYang:
I reviewed and approved this PR.

PR #1205, opened 2026-06-02
Metal GPU + CoreML/ANE support for transformer nets (SiLU, GQA, learnable RoPE)

ChinChangYang:
## Summary

Adds Metal GPU and CoreML/ANE (Apple Neural Engine) inference support for the v15+ transformer trunk
architectures (e.g. `b10c384h6nbttflrs`), bringing the macOS backends to parity with the new
transformer features, and fixes a convnet performance regression and a Metal GPU memory leak found
along the way.

This targets the macOS Metal 4 backend and the bundled `katagocoreml` MIL converter only; it does not
touch the CUDA / TensorRT / OpenCL / Eigen backends.

## What's included

- **Transformer trunk support** on both the Metal GPU and CoreML/ANE paths:
  - Attention blocks, including **Grouped-Query Attention (GQA)**
  - **Learnable RoPE** positional encoding
  - **SiLU** activation (previously dropped in the CoreML value / policy / meta heads)
  - RMSNorm / batchnorm tips
- **Metal GPU crash fix**: GQA attention overflowed an `NDArray` dimension (`INT_MAX`); reshaped to avoid it.
- **Metal GPU memory leak fix** (`1e885b72`): the GPU-mode inference path (`MPSGraphModelHandle.apply`)
  allocated a command buffer, the four MPSNDArray / MPSGraphTensorData input feeds, and the encode result
  tensors per batch — all autoreleased Obj-C objects — with no `autoreleasepool`. NN-eval server threads
  are long-lived and never exit during a multi-game run, so these accumulated unboundedly: ~1 GB leaked per
  game (observed 93 GB footprint + 20 GB swap after ~76 games, causing self-inflicted jetsam kills).
  Wrapping the `apply()` body in `autoreleasepool {}` (mirroring the CoreML `apply()` path) drains each
  batch's Metal objects immediately — ~1 GB/game → ~11 MB/game (bounded nn-cache warmup), footprint flat
  across a long run. Affects all long-running GPU-mode processes (selfplay, benchmark, analysis).
- **CoreML/ANE FP16 accuracy** via precision tiers: the ANE is FP16-only and accumulates in FP16, so the
  accuracy-sensitive ops (non-spatial pooling, RMSNorm, heads, and — for narrow trunks — the whole graph)
  are escalated to FP32 while attention stays FP16. Mirrors the TensorRT precision split.
- **Convnet regression fix** (`d052d2a1`): the precision tiers were originally keyed on trunk width, which
  also caught plain production convnets (b18c384 / b28c512 / b60c320) and forced FP32 ops off the FP16-only
  ANE (~2.6× slower). The escalations are now gated on **actual transformer-block presence** (recursing into
  nested-bottleneck blocks), so convnets run pure FP16 on the ANE (identical to prior behavior) and
  transformer graphs are byte-for-byte unchanged.

## Verification (macOS, Apple Silicon)

- **CoreML/ANE FP16 path** (`metalDeviceToUseThread0=100, metalUseFP16=true`): `cpp/rungpuerrortest.sh`
  run two-phase (Eigen FP32 references → Metal compare): **32/32 CoreML-supported configs pass** across
  board sizes 9 / 13 / 19 / 10×14 / rectangle, including the rectbuffer and weird-settings cases. Worst
  ANE-FP16-vs-Eigen winrate max = 1.13% (threshold 5.0%); `fp32`-vs-reference ≈ 0.0005% confirms valid
  references. Models exercised: b18c384nbt ×2, b28c512nbt, b18c384nbt-humanv0, b5c192nbt, g170e-b10c128,
  and 3 transformer test nets (fixed-RoPE / learnable-RoPE+SiLU / GQA).
- **Metal GPU (MPSGraph, FP32) path** — the default device (`gpuIdx=0` = `METAL_MUX_GPU` → MPSGraph; no
  `metalDeviceToUseThread0=100` override): `testgpuerror` against the same Eigen FP32 references for all 3
  transformer test nets, **13/13 configs pass** across board sizes 9 / 13 / 19 / rectangle plus the
  rectbuffer case. Worst MPSGraph-vs-Eigen winrate max ≈ 0.0006% (threshold 5.0%); `fp32`-vs-reference
  matches, i.e. pure FP32-vs-FP32 numerical noise. Covers the fixed-RoPE + RMSNorm-tip (b7c96h3tfrs),
  learnable-RoPE + SiLU + spatial-RMSNorm-tip (b4c256h4nbttflrs), and GQA + ReLU + batchnorm-tip
  (b7c96h6kv3qk32v16tflrs) architectures. The MPSGraph path is FP32-only by design.
- **Memory leak fix** is numerically inert (an `autoreleasepool` changes only object lifetime, not output):
  on the post-fix build, `testgpuerror` Metal-vs-Eigen is unchanged for the transformer + convnet nets
  (max winrate err ~0.0005%), and the GPU-mode `apply()` path runs cleanly. The leak reduction itself was
  validated separately (~1 GB/game → ~11 MB/game, footprint flat at ~1.1 GB across a long run).
- `./katago runtests` and `./katago runnnlayertests` pass.
- b18c384nbt ANE throughput restored by the regression fix: **160 → 424 visits/s**.

## Notes

- Ready for review of the macOS Metal / CoreML transformer support.
- Pre-v8 KataGo model versions are not supported by the CoreML converter (`MIN_SUPPORTED_VERSION = 8`); such
  nets are expected to run on the GPU / other backends.


ChinChangYang:
I have reviewed and approved this.

lightvector:
Thanks! Merged, with some additional edits improving the error handling and robustness and adding more checks, and also performing one of the todos regarding the value head that empirically shrinks average errors quite a bit.
8c1a9ad2
3d2cdca2
9c26f4ac

ChinChangYang:
Reviewed the three commits. They look good to me.

PR #1234, opened 2026-08-12
Backends to merge

lightvector:
Candidate branch to merge to master. Combines https://github.com/lightvector/KataGo/pull/1222 ONNX backend from @seniorfish and https://github.com/lightvector/KataGo/pull/1188 ROCm backend from @Looong01 along with a bunch of followup fixes and refactors I added on top of each backend from local review and testing and a few other features I also worked on recently.

Will go into master if everyone confirms that it's working well.

PR #1237, opened 2026-08-14
[ROCm] Fix slowdown and Compiling.md: 1x1 convs as hipBLAS GEMM, NCHW convnet default, MIOpen naive-solver detection, ROCm 7.14 install docs

Looong01:
Follow-up to #1234 (the unified CUDA/ROCm backend). While benchmarking the ROCm path of the merged tip on a Radeon RX 7900 XTX (gfx1100, ROCm 7.14, MIOpen 3.5.2), I found that MIOpen's NHWC convolution solver coverage is much weaker than cuDNN's, and that a few backend defaults inherited from the CUDA side cost real performance on MIOpen. This PR fixes those, and updates the ROCm install docs to the official ROCm 7.14 guide.

## Problems

1. **1x1 NHWC FP16 convs are slow through MIOpen.** For the transformer head convs, MIOpen picks `ConvHipImplicitGemmGroupFwdXdlops` (CK grouped-conv WMMA kernels) - measured ~2-4x slower per call than an equivalent GEMM, with far more dispatch overhead (thousands of extra conv dispatches per search batch).
2. **FP32 1x1 NHWC convs (e.g. `policy_head.conv2p`) can fall all the way back to `ConvDirectNaiveConvFwd`** - MIOpen's naive direct fallback.
3. **`rocmUseNHWC=Auto` enabled NHWC for convnets on CDNA**, but MIOpen's NCHW solver coverage (Winograd / implicit GEMM / GEMM) is its most mature path on every supported arch, including CDNA.
4. **Docs:** the ROCm install instructions still referenced TheRock directly; ROCm 7.14 has official packages (apt on Linux, tarball on Windows).

## Changes

- **`rocmUse1x1Matmul`** (new config, mirrors `cudaUse1x1Matmul`, Auto = on): 1x1 NHWC convs run as hipBLAS GEMM. FP16 uses `hipblasGemmEx` with FP32 compute for accuracy (same reasoning as `MatMulLayer`); FP32 uses it too, since unlike cuDNN, MIOpen's NHWC path is weak in both precisions.
- **`rocmUseNHWC=Auto` now resolves to NCHW for convnets on all archs.** Transformer models are unchanged - they still force NHWC (the attention/RoPE/FFN implementation requires it), and their 1x1 convs are now GEMMs anyway.
- **MIOpen naive-fallback detection at model load.** MIOpen stamps its naive fallback solution with a sentinel time of 1000ms (registered solver id 55 as a secondary check - note the id is not public-stable API, hence two signals). If an NHWC conv's chosen solution is naive, the layer is re-expressed as NCHW with per-call transposes; a log line reports the fallback and the re-expression, so coverage gaps are visible instead of silent.
- **Compiling.md**: ROCm install sections (Linux + Windows) rewritten to follow <https://rocm.docs.amd.com/en/latest/install/rocm.html> - official apt packages on Linux (`amdrocm7.14 amdrocm-core-dev7.14 ...`, with the 7.2.4-or-older uninstall note and `amd-smi` verification) and the official tarball layout on Windows.
- Config docs for `rocmUse1x1Matmul` / `rocmUseNHWC` updated in `gtp_example.cfg` and `analysis_example.cfg`.

Scope note: all changes are inside the ROCm (`KATAGO_GPU_HIP`) sections or ROCm config parsing; the CUDA path is untouched.

## Measurements

RX 7900 XTX (gfx1100), ROCm 7.14.0, MIOpen 3.5.2, CK 1.2.0, 19x19, FP16, `benchmark` at 800 visits, best visits/s across the thread sweep:

| Model | master tip | this PR | delta |
|---|---|---|---|
| b10c384h6nbttflrs | 2649 | 2835 | +7.0% |
| b10c512h8nbt3tflrs-fson-silu-rsnh | 1506 | 1551 | +3.0% |
| b11c768h12nbt3tflrs-fson-silu | 780 | 830 | +6.4% |
| kata1-zhizi-b40c768nbt-s11272M-d5935M | 354 | 346 | neutral (within noise) |

Correctness: FP16 `kata-raw-nn` outputs on a fixed position checked against the Eigen CPU backend as reference - whiteWin/whiteLead agree to ~1e-3 on b10c384, b10c512, b11 and the b40 convnet (e.g. b11: 0.7498 vs 0.7499).

## Caveats

- The naive-fallback detection relies on MIOpen's sentinel time (1000ms), which is stable behavior, and secondarily on the registered solver id (55 for `ConvDirectNaiveConvFwd` in MIOpen 3.5.x), which is not guaranteed stable across MIOpen versions. If a future MIOpen renumbers it, the time sentinel still catches fallbacks.
- The NCHW re-expression adds two transposes per call; it only engages when MIOpen's NHWC options are down to naive, where even naive-plus-nothing is worse. On gfx1100 the affected layers are tiny head convs, so the transpose cost is negligible.

### DIFF
--- a/Compiling.md
+++ b/Compiling.md
@@ -158,7 +158,7 @@ As also mentioned in the instructions below but repeated here for visibility, if
       * The ROCm (MIOpen) backend supports Windows via [AMD TheRock](https://github.com/ROCm/TheRock) (tested with TheRock 7.13 / ROCm 7.13, RX 7900 XTX / gfx1100), including transformer/attention models (model version 17+) and the optional Composable Kernel (CK) fused-attention fast path.
       * **Prerequisites:**
          * Download [AMD TheRock](https://github.com/ROCm/TheRock) and extract it to e.g. `C:\TheRock\build`, adjusting the paths below if you extract elsewhere.
-         * Install **Visual Studio Build Tools or Community** with the "Desktop development with C++" workload, for the MSVC toolchain and Windows SDK the HIP compiler needs. A **v143 toolset (MSVC 14.3x or 14.4x)** must be among the installed toolsets - newer toolsets alone (14.5x+) are not accepted by the HIP clang compatibility check. If more than one is installed side by side, `CMakeLists.txt` automatically probes them at configure time and picks a compatible one itself (see "Fully automatic" below), no manual toolset selection needed.
+         * Install **Visual Studio Build Tools or Community** with the "Desktop development with C++" workload, for the MSVC toolchain and Windows SDK the HIP compiler needs. If several toolsets are installed side by side, `CMakeLists.txt` probes them at configure time, newest first, and picks the first one the HIP compiler actually accepts (see "Fully automatic" below), no manual toolset selection needed. A very new MSVC STL can be ahead of TheRock's bundled clang, so if configure reports that no installed toolset works, add an older one such as **MSVC v143 (14.3x/14.4x)** from the Visual Studio Installer's "Individual Components" tab.
          * Install [Ninja](https://ninja-build.org) build tool: `winget install Ninja-build.Ninja`.
          * Set the following **system environment variables** (via System Properties -> Advanced -> Environment Variables):
            ```
@@ -186,9 +186,9 @@ As also mentioned in the instructions below but repeated here for visibility, if
         Windows-specific setup automatically at configure/build time:
          * **MSVC toolset selection:** if more than one MSVC toolset is installed side by side, a
            newer one can conflict with TheRock's bundled clang (newer MSVC STL headers are not yet
-           compatible with it). `CMakeLists.txt` finds the installed v143-family toolsets via
-           `vswhere` and probes each with a real compile until one works, with no user action
-           needed.
+           compatible with it). `CMakeLists.txt` enumerates every installed toolset via
+           `vswhere` and probes them newest-first with a real compile until one works, with
+           no user action needed.
          * **zlib:** TheRock's Windows package ships `zlib.h` but (as of 7.13) no longer ships a
            linkable `.lib`. `CMakeLists.txt` automatically bootstraps a local
            [vcpkg](https://github.com/microsoft/vcpkg) clone under `<build dir>/deps/vcpkg` (this
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -137,15 +137,29 @@ function(katago_win_apply_vcvarsall vcvarsall_bat vcvars_ver)
   if(NOT _rc EQUAL 0 OR NOT EXISTS "${_out_file}")
     return()
   endif()
-  # Read line-by-line via file(STRINGS) rather than splitting the raw output ourselves - INCLUDE/
-  # LIB/PATH are themselves semicolon-separated, which would corrupt a naive string(REPLACE "\n" ";"
-  # ...) split. file(STRINGS) keeps each real line, with its embedded semicolons, as one entry.
-  file(STRINGS "${_out_file}" _env_lines)
-  foreach(_line IN LISTS _env_lines)
+  # Walk the dump one line at a time with plain string operations, not by iterating a cmake list
+  # from file(STRINGS). Two properties of this output make list handling wrong: INCLUDE, LIB, and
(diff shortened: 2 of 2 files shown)


lightvector:
Question: this PR deletes this code:
```
  if(context->useNHWCMode == enabled_t::Auto && useFP16 && isNhwcFp16PreferredArch(prop))
    useNHWC = true;
```

which used NHWC on some CDNA architectures. Is this deletion something that your benchmarks actually tested, or did you test RDNA only and are only guessing about this deletion? My testing found that this is a lot worse on MI300X (which I think is gfx942), on conv nets from 1.3x to 2x slower depending on the model and thread settings.

Looong01:
> Question: this PR deletes this code:
> 
> ```
>   if(context->useNHWCMode == enabled_t::Auto && useFP16 && isNhwcFp16PreferredArch(prop))
>     useNHWC = true;
> ```
> 
> which used NHWC on some CDNA architectures. Is this deletion something that your benchmarks actually tested, or did you test RDNA only and are only guessing about this deletion? My testing found that this is a lot worse on MI300X (which I think is gfx942), on conv nets from 1.3x to 2x slower depending on the model and thread settings.

That deletion is based on **RDNA3 (gfx1100, RX 7900 XTX) testing only**. I do not have a CDNA/MI300X (gfx942) machine, so I cannot claim it is optimal there without data.

What we observed on gfx1100 is that MIOpen's NHWC path frequently falls back to the `ConvDirectNaiveConv*` solver, which is dramatically slower than the NCHW path. The b11c768h12 model showed a ~15% regression with NHWC auto-enabled, and the smaller conv nets were worse as well. Removing the auto-enable avoids this regression on RDNA3.

Your finding that NHWC is also worse on MI300X (1.3–2× slower on conv nets) is consistent with our observation. **Do u mean this happens after this PR or before this PR? If it is before this PR**, then the conservative default is the same: do not auto-enable NHWC on AMD GPUs through MIOpen. If someone later produces CDNA-specific evidence that NHWC is faster on a specific MI300X/MI325X configuration, that can be revisited with a more targeted arch guard.

lightvector:
It happens after the PR. The specific lines I cited appear to be a deletion that cannot possibly affect gfx1100 because isNhwcFp16PreferredArch already is false on gfx1100. Why did you delete it on the basis of gfx1100 testing when such testing could not possibly exercise the effect of the deletion? And yes, deleting those lines did indeed prove to be harmful on gfx942.

lightvector:
(Sorry, closing the PR was my laptop touchpad misfiring as I was moving the mouse)

lightvector:
> Your finding that NHWC is also worse on MI300X (1.3–2× slower on conv nets) is consistent with our observation. Do u mean this happens after this PR or before this PR?

Ah I think I understand the confusion. To be clear: NHWC is 1.3x to 2x *better* on MI300X. The deletion of those lines in the PR, disabling NHWC, made it 1.3x to 2x worse. And I was asking about them because their deletion only affects some of the gfx9... gpus, which wouldn't have been testable in gfx1100 benchmarks, so I was wondering why the deletion was made.

Looong01:
@lightvector You're right — that deletion was a mistake.

I removed the CDNA auto-NHWC branch based only on RDNA3 (gfx1100) testing, where that branch already never fired, so I had no evidence to generalize it to all AMD architectures. I incorrectly treated a local RDNA3 observation as universal and it predictably hurt gfx942.

I restored it and broadened the rule: `rocmUseNHWC=auto` now enables NHWC for the entire `gfx9*` family. The implementation uses `std::string_view` so the prefix length is not hard-coded:

```cpp
static bool isNhwcPreferredArch(const hipDeviceProp_t& prop) {
::string arch(prop.gcnArchName);
  size_t colonPos = arch.find(':');
  if(colonPos != std::string::npos)
    arch = arch.substr(0, colonPos);
  static constexpr std::string_view prefix = "gfx9";
  return arch.compare(0, prefix.size(), prefix) == 0;
}

Looong01:
Finished

Looong01:
@lightvector I make a little commit here: https://github.com/lightvector/KataGo/pull/1237/commits/2f1153f791732365d5aafc680acd44a3d7af051f

Just Update CMakeLists about checking MSVC toolsets version logic for ROCm backend building on Windows. I think u can add this to our latest Update.

lightvector:
Added.

PR #1241, opened 2026-08-24
Fix OpenCL backend NCHW WMMA layout identifier typo

dhkim92-dev:
# Summary
Errata fix on OpenCL Backend
This change corrects a typographical error in the OpenCL WMMA path flags name 

**usingHGemmWmmaNHCW** is renamed to **usingHGemmWmmaNCHW**

The implementation already uses NCHW layout, so this is a only naming-only correction and doesn't change any logics.


lightvector:
Thanks!

PR #1250, opened 2026-09-03
Fix model download resume with RFC-compliant Range requests

Carton:
## Summary

- Send RFC 7233-compliant `Range: bytes=start-end` headers when resuming model downloads.
- Validate response status and `Content-Range` before appending data.
- Discard untrusted partial data and fall back to a clean download when Range is ignored or malformed.
- Keep interrupted full downloads on the normal retry budget when Range is unavailable.

## Validation

- Built the `katago` executable successfully.
- Ran the built-in test suite successfully.
- Verified three fresh-base-dir downloads with controlled mid-stream connection drops; each resumed from the partial offset, completed downloading full file, and passed SHA-256/model loading validation.

### DIFF
--- a/cpp/distributed/client.cpp
+++ b/cpp/distributed/client.cpp
@@ -207,11 +207,15 @@ static httplib::Result oneShotDownload(
   const Url& proxyUrl,
   size_t startByte, //inclusive
   size_t endByte, //inclusive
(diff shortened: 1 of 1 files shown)


lightvector:
Thanks for noticing this!

PR #1253, opened 2026-09-10
Fix contribute watchOngoingGameInFile for non-ASCII paths on Windows

Hoigin:
## Summary

On Windows, if the path in `watchOngoingGameFileName` contains non-ASCII characters, like `D:\Go\跑谱\watchgame.txt`, KataGo fails to open the watch file, and the whole contribute process terminates a few seconds after startup with:
```
ERROR: game loop loop thread failed: Could not open file: D:\Go\跑谱\watchgame.txt
```
The root cause is that the Windows branch opens the file with the narrow-char `fopen_s`, which interprets the UTF-8 path bytes in the ANSI codepage, so the path no longer matches the real directory (a TODO in contribute.cpp already notes that non-ascii paths are unhandled).

This PR converts the path to wide chars and opens it with `_wfopen`, the same pattern already used in `homedata.cpp` and `mainargs.cpp`.

## Testing

Verified on Windows 11 (Chinese locale, GBK codepage), KataGo v1.18.2 + this change, MSVC build with `-DBUILD_DISTRIBUTED=1`:

- A path containing a non-ASCII directory (`D:\Go\跑谱\watchgame.txt`) now creates the file at the correct path. ASCII-only paths behave as before.
- Invalid UTF-8 in the config value produces a clear StringError.

### DIFF
--- a/cpp/command/contribute.cpp
+++ b/cpp/command/contribute.cpp
@@ -22,6 +22,15 @@
 #include "../command/commandline.h"
 #include "../main.h"
 
+#if __MINGW32__
+// `std::wstring_convert` needs explicit including in case of MINGW
+#include <locale>
+#endif
+
+#ifdef OS_IS_WINDOWS
+#include <codecvt>
+#endif
+
 #ifndef BUILD_DISTRIBUTED
 
 int MainCmds::contribute(const std::vector<std::string>& args) {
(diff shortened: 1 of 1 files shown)


lightvector:
Thanks for reporting! Fixed in a slightly different way, in https://github.com/lightvector/KataGo/commit/9bb7efc39b2b432133c964ece6f8fb969441fdfc

Hoigin:
Verified the alternative fix, closing this in favor of it. Thanks for the great project!

